PR #1036, opened 2014-12-27
Set Content-Type: application/json

msabramo:
for a bunch of endpoints that output JSON.

This makes output nicer when accessing via a tool, such as HTTPie


msabramo:
Added check for `Content-Type` to several tests.

~~~go
c.Assert(
    recorder.Header().Get("Content-Type"),
    gocheck.Equals,
    "application/json")
~~~

PR #1038, opened 2014-12-27
Log to stderr

msabramo:
in addition to syslog or file

Some folks don't use syslog. They may want to run `tsr` using a process manager like supervisord that manages log files for output to stdout/stderr.

Also it is much easier to look at the console when running `tsr api` in the terminal when developing or debugging.

E.g.:

    $ ~/go/bin/tsr api
    ...
    2014-12-27T07:13:12.851705809-08:00 GET /auth/scheme 200 in 0.519067ms
    2014/12/27 07:13:16 DEBUG: Ignored invalid token for /users: Invalid token
    2014/12/27 07:13:16 DEBUG: Failed to create user in the git server: Failed to connect to Gandalf server, it's probably down.

### DIFF
--- a/log/multi_logger.go
+++ b/log/multi_logger.go
@@ -1,0 +1,56 @@
+// Copyright 2014 tsuru authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+package log
+
+import (
+	"log"
+)
+
+func NewMultiLogger(loggers ...*Logger) Logger {
+	return &multiLogger{loggers: loggers}
+}
+
+type multiLogger struct {
+	loggers []*Logger
+}
+
+func (m *multiLogger) Debug(format string) {
+	for _, logger := range m.loggers {
+		(*logger).Debug(format)
+	}
+}
+
+func (m *multiLogger) Error(format string) {
+	for _, logger := range m.loggers {
+		(*logger).Debug(format)
+	}
+}
+
+func (m *multiLogger) Fatal(format string) {
+	for _, logger := range m.loggers {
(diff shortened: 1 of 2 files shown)


msabramo:
What's the drawback if it always logs to stderr?

log/multi_logger.go
```diff
@@ -0,0 +1,58 @@
+// Copyright 2014 tsuru authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+package log
+
+import (
+	// 	"fmt"
```
andrewsmedina: Can you remove these comments?

andrewsmedina:
I dont know if have some drawbacks, but I thinking if there is a case that we dont want send the log to file/syslog + stderr.
Another thing, we should send all log data to stderr or just send the error level is enough?

msabramo:
Yeah. I don't know either. In Python, logging is very configurable as you can have multiple loggers with different handlers. The golang stdlib log module seems less flexible. 

I was looking at some alternatives like logrus and log15 - these seem nice but that's a big change so I'd leave that up to you guys.

log/multi_logger.go
msabramo: Done.

andrewsmedina:
@msabramo can you add tests for `multiLogger`?

fsouza:
@msabramo can I take over this pull request and provide the needed fixes and tests?

msabramo:
@fsouza: Absolutely. I haven't had as much time to work on tsuru so go for it!

fsouza:
This is now merged. I thought github would close this automatically.

PR #1046, opened 2015-01-01
.travis.yml: Allow go tip to fail

msabramo:
Because it might not always be stable.

See https://github.com/travis-ci/travis-ci/issues/2944:

> It's probably best to keep tip as an allowed_failures entry if you're
> concerned about this causing disruptions.


andrewsmedina:
@msabramo thank you!

PR #1047, opened 2015-01-01
Make `tsr token` include newline

msabramo:
Make sure output ends with a newline, as this is a UNIX convention to always have a newline at the end of text.


andrewsmedina:
@msabramo thank you!

PR #1049, opened 2015-01-01
travis: wait for MongDB.

andrewsmedina:


msabramo:
Ah, same as what I did after you pointed out `before_script`. I think you got lucky here as it seems to fail intermittently.

.travis.yml
```diff
@@ -15,6 +15,8 @@ install:
   - export PATH="$HOME/gopath/bin:$PATH"
   - echo http://localhost > $HOME/.tsuru_target
   - make get GO_EXTRAFLAGS=-x
+before_script:
+  - nc -zvv localhost 27017; while [[ $! -ne 0 ]]; do nc -zvv localhost 27017; done
```
msabramo: Maybe add `echo` and `sleep` so that we can see when it retries and how long it takes?

PR #1048, opened 2015-01-01
.travis.yml: Add sleep 15

msabramo:
To see if it prevents problems connecting to MongoDB, as suggested at http://docs.travis-ci.com/user/database-setup/#MongoDB-may-not-be-immediately-accepting-connections

PR #1050, opened 2015-01-01
.travis.yml: Use travis_retry for `make test`

msabramo:
Hoping that this might help Travis CI tests to pass even when there are sporadic failures relating to MongoDB (e.g.: #1048, #1049)


cezarsa:
Let's give it a try, thanks!

PR #1051, opened 2015-01-01
.travis.yml: wait for Redis port 6379

msabramo:
And add `echo` and `sleep` to `while` loops for retrying.


andrewsmedina:
@msabramo thank you!

PR #1057, opened 2015-01-04
.travis.yml: Move "services" earlier in file

msabramo:
probably won't help (with intermittent test failures related to connecting to MongoDB) but probably also won't hurt :)


cezarsa:
This won't help as yaml entries are parsed as an unordered map.
From the latest travis builds we can see that the problem is really related to processes being killed by lack of available memory.

Issue #1045, opened 2014-12-31
Travis CI tests failing intermittently


msabramo:
Most pull requests fail in either the Go 1.4 or Go tip sub job, but there is no obvious pattern, so I am guessing that the tests fail intermittently. Haven't looked in detail though.

andrewsmedina:
It looks like some problem with Travis CI + MongoDB. In drone.io we don't have this problem: https://drone.io/github.com/tsuru/tsuru

msabramo:
Some ideas at https://github.com/travis-ci/travis-ci/issues/3075#issuecomment-68629130

Try adding `sudo dmesg` to `after_failure` to see if OOMs are happening.

cezarsa:
Good call @msabramo. It's definitely related to OOMs as we can see in https://travis-ci.org/tsuru/tsuru/jobs/46092093#L2702
```
[10314713.731361] Out of memory in UB 197947879: OOM killed process 1852 (mongod) score 0 vm:3174284kB, rss:70808kB, swap:2168kB
[10314821.568655] Out of memory in UB 197947879: OOM killed process 8832 (mysqld) score 0 vm:908892kB, rss:107020kB, swap:0kB
[10314821.635923] Out of memory in UB 197947879: OOM killed process 9137 (sh) score 0 vm:4348kB, rss:672kB, swap:0kB
[10314821.636752] Out of memory in UB 197947879: OOM killed process 9141 (sleep) score 0 vm:4252kB, rss:528kB, swap:0kB
``` 
However I'm not sure about how to solve this. We could try killing some processes before starting the tests. We don't need mysqld running. Also setting GOMAXPROCS=1 might help.

msabramo:
There are a few ideas for solving OOM failures over at https://github.com/travis-ci/travis-ci/issues/3075#issuecomment-68629130

I wonder if we can add a `ps` or `top` that sorts by memory to see what processes are using the most memory.

msabramo:
Maybe something like this?

```
vagrant@vagrant-ubuntu-trusty-64:~$ top -b -n 1 -o RES | head -n 15
top - 05:59:04 up 6 min,  1 user,  load average: 0.01, 0.18, 0.13
Tasks: 103 total,   1 running, 102 sleeping,   0 stopped,   0 zombie
%Cpu(s):  3.1 us,  1.8 sy,  0.3 ni, 94.5 id,  0.2 wa,  0.0 hi,  0.1 si,  0.0 st
[...]

PR #1060, opened 2015-01-07
.travis.yml: ++memory info in after_failure

msabramo:
Add `free` and `top -b -n 1 -o RES | head -n 15` so we can try to see where all of the memory is going.

Refs: #1045

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,12 @@ install:
   - echo http://localhost > $HOME/.tsuru_target
   - make get GO_EXTRAFLAGS=-x
 before_script:
+  # Check baseline memory usage; for compare with after_failure when OOM occurs
+  - free
+  - vmstat
+  - ps aux --sort -rss | head -n 10
+  - mongostat -n 1
+  # Make sure MongoDB and Redis are available
   - nc -zvv localhost 27017; out=$?; while [[ $out -ne 0 ]]; do echo "Retry hit port 27017..."; nc -zvv localhost 27017; out=$?; sleep 1; done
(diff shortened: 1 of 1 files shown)


cezarsa:
Seems like travis doesn't support "-o", could you try something like: `top -b -n 1 | grep "^ *[0-9]" | sort -rn -k 6`
Also, it would be useful to run the same commands at the end of the `before_script` section so we can see what are our initial conditions.

msabramo:
@cezarsa: Just updated it to use `ps -aux --sort -rss | head -n 10`. That seems a little nicer only because we don't have to use `grep` and then lose the column headers.

msabramo:
From looking at the last travis ci job, it looks like maybe the tests are just putting too much data into MongoDB?

```
1.05s$ mongostat -n 1
connected to: 127.0.0.1
insert  query update delete getmore command flushes mapped  vsize    res faults                   locked db idx miss %     qr|qw   ar|aw  netIn netOut  conn       time 
    *0     *0     *0     *0       0     1|0       0  2.58g  5.38g    24m      0 tsuru_api_handler_test:0.0%          0       0|0     0|0    62b     9k     1   19:03:28 
Done. Your build exited with 1.
```

cezarsa:
We don't need neither postgresql nor mysql, stopping them in before_script may help:
```
- sudo /etc/init.d/postgresql stop
- sudo /etc/init.d/mysql stop
```

msabramo:
@cezarsa: OK, see https://github.com/tsuru/tsuru/pull/1062

But I think MongoDB is using far more memory than postgresql and mysql. Perhaps we need to be clearing out the data that gets stored in there in between tests?

msabramo:
@cezarsa, @andrewsmedina: Want to merge this so we get this info on every build and see if we're making progress reducing memory usage?

PR #1062, opened 2015-01-07
.travis.yml: Stop postgresql and mysql

msabramo:
to save some memory and maybe lessen chance of OOMs on Travis

(although I think MongoDB is using a lot more memory than these two are).


msabramo:
Also see https://github.com/travis-ci/travis-ci/issues/2904#issuecomment-148485284

ribasushi:
@msabramo Ignore me - I didn't read the commit date on the "referenced" part of the issue. FAIL.

PR #1066, opened 2015-01-08
Display info about router on startup

msabramo:
Example output:

    [marca@marca-mac2 tsuru]$ make run-tsr-api
    build/tsr api
    Opening config file: /etc/tsuru/tsuru.conf
    Done reading config file: /etc/tsuru/tsuru.conf
    Using mongodb database "tsurudb" from the server "127.0.0.1:27017".
    Using hipache router "10.128.40.70.xip.io".  <--------------------- This is the new information
    Using "docker" provisioner.
    Using "native" auth scheme.
    tsuru HTTP server listening at 0.0.0.0:8080...

Mainly I am hoping that this helps people to understand the various pieces of tsuru better and what depends on what, because this is what I struggled with the most when I first started playing with it.

### DIFF
--- a/router/galeb/router.go
+++ b/router/galeb/router.go
@@ -310,3 +310,15 @@ func (r *galebRouter) Routes(name string) ([]string, error) {
 	}
 	return hosts, nil
 }
+
+func (r galebRouter) StartupMessage() (string, error) {
+	domain, err := config.GetString(r.prefix + ":domain")
+	if err != nil {
+		return "", err
+	}
+	apiUrl, err := config.GetString(r.prefix + ":api-url")
+	if err != nil {
+		return "", err
+	}
+	return fmt.Sprintf("Using galeb router %q with API URL %q.\n", domain, apiUrl), nil
+}
--- a/api/server.go
+++ b/api/server.go
@@ -18,6 +18,7 @@ import (
 	"github.com/tsuru/tsuru/db"
 	"github.com/tsuru/tsuru/log"
 	"github.com/tsuru/tsuru/provision"
+	"github.com/tsuru/tsuru/router"
 )
 
 func getProvisioner() (string, error) {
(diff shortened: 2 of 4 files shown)


msabramo:
I wonder if it's desirable and possible to print the Docker nodes that tsuru knows about?

It seems useful to know what your Docker nodes are, especially since I've seen at least a few times on gitter/IRC, people were puzzled when they got an error about having a node in the config file that was already registered.

msabramo:
Also see https://github.com/tsuru/tsuru/pull/1071

andrewsmedina:
In the 0.10.0 `tsuru` will support more than one router. It is already working on master. So the `docker:router` is the default router and you can set custom routers by `plan`.

So, the "hipache:domain" can be empty. For now, I believe that you can read the "docker:router" if the router is hipache you shoud print the "hipache:domain" value if not you can ignore it or another way is print the "galeb:domain".

msabramo:
I think this can probably be done more in the style of what @cezarsa suggested in #1071 -- instead of reading stuff out of the config in `app/server.go`, I could have it tell the router to output any interesting information about itself...

msabramo:
Updated this to be more like #1071 and now it's more general and can print info about hipache or galeb routers.

msabramo:
Just updated again, applying the same comments that @cezarsa suggested in #1071.

router/galeb/router.go
```diff
@@ -310,3 +311,16 @@ func (r *galebRouter) Routes(name string) ([]string, error) {
 	}
 	return hosts, nil
 }
+
+func (r galebRouter) FprintStartupMessage(w io.Writer) error {
```
cezarsa: I think you missed changing the method signature here.

msabramo: Indeed! Just updated. Thanks!

cezarsa:
Thanks!

PR #1067, opened 2015-01-09
Change sample port in docs from 8888 to 8100

msabramo:
because I found evidence of the old admin-listen which was on port 8888
and now when I grep for 'admin-listen|:8888', I get false positives from
the docs.

### DIFF
--- a/docs/using/go.rst
+++ b/docs/using/go.rst
@@ -80,12 +80,13 @@ A simple web application in go `main.go`:
     import (
         "fmt"
         "net/http"
+        "os"
     )
 
     func main() {
         http.HandleFunc("/", hello)
         fmt.Println("listening...")
-        err := http.ListenAndServe(":8888", nil)
+        err := http.ListenAndServe(":" + os.Getenv("PORT"), nil)
         if err != nil {
             panic(err)
         }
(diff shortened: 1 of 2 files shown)


andrewsmedina:
@msabramo the best practice here is to get the port from the "PORT" environment variable.

msabramo:
Good point!

msabramo:
@andrewsmedina: Updated to use `PORT`

docs/using/go.rst
```diff
@@ -80,12 +80,13 @@ A simple web application in go `main.go`:
...
         "fmt"
         "net/http"
+        "os"
     )
 
     func main() {
         http.HandleFunc("/", hello)
         fmt.Println("listening...")
-        err := http.ListenAndServe(":8888", nil)
+        err := http.ListenAndServe(os.Getenv("PORT"), nil)
```
andrewsmedina: the ":" is missing

msabramo: oops

msabramo:
```
❯ git push -f msabramo change_sample_port_in_docs_from_8888_to_8100
Counting objects: 5, done.
Delta compression using up to 8 threads.
Compressing objects: 100% (5/5), done.
Writing objects: 100% (5/5), 549 bytes | 0 bytes/s, done.
Total 5 (delta 4), reused 0 (delta 0)
To git@github.com:msabramo/tsuru.git
 + bf83b60...0344355 change_sample_port_in_docs_from_8888_to_8100 -> change_sample_port_in_docs_from_8888_to_8100 (forced update)
```

andrewsmedina:
@msabramo thank you!

PR #1068, opened 2015-01-09
Remove "admin-listen" stuff from configs

msabramo:
admin-listen (port 8888) is not used anymore in tsuru, according to @andrewsmedina


andrewsmedina:
@msabramo thank you!

PR #1071, opened 2015-01-09
Show Docker nodes at startup

msabramo:
This is my stab at making tsuru display the list of nodes in the Docker cluster when tsuru starts up.

It seems useful to know what your Docker nodes are, especially since I've seen at least a few times on gitter/IRC, people were puzzled when they got an error about having a node in the config file that was already registered.

```
vagrant@vagrant-ubuntu-trusty-64:~/go/src/github.com/tsuru/tsuru$ rm -rf build; make run-tsr-api
godep go build -o build/tsr cmd/tsr/*.go
build/tsr api
Opening config file: /etc/tsuru/tsuru.conf
Done reading config file: /etc/tsuru/tsuru.conf
Using mongodb database "tsurudb" from the server "127.0.0.1:27017".
Using "docker" provisioner.
nodeList = [{http://127.0.0.1:2375 {0001-01-01 00:00:00 +0000 UTC false} map[LastSuccess:2015-01-04T18:22:22Z pool:theonepool]}]
Using "native" auth scheme.
tsuru HTTP server listening at 0.0.0.0:8080...
```

### DIFF
--- a/api/server.go
+++ b/api/server.go
@@ -230,6 +230,12 @@ func RunServer(dry bool) http.Handler {
 				fatal(err)
 			}
 		}
+		if messageProvisioner, ok := app.Provisioner.(provision.MessageProvisioner); ok {
+			startupMessage, err := messageProvisioner.StartupMessage()
+			if err == nil && startupMessage != "" {
+				fmt.Print(startupMessage)
+			}
+		}
 		scheme, err := getAuthScheme()
 		if err != nil {
 			fmt.Printf("Warning: configuration didn't declare a auth:scheme, using default scheme.\n")
--- a/provision/docker/provisioner.go
+++ b/provision/docker/provisioner.go
@@ -40,7 +40,19 @@ func getRouterForApp(app provision.App) (router.Router, error) {
 	return router.Get(routerName)
 }
 
-type dockerProvisioner struct{}
+type dockerProvisioner struct {}
+
+func (p *dockerProvisioner) StartupMessage() (string, error) {
+	nodeList, err := dockerCluster().UnfilteredNodes()
+	if err != nil {
+		return "", err
+	}
+	out := "Docker provisioner reports the following nodes:\n"
+	for _, node := range nodeList {
+		out += fmt.Sprintf("    Docker node: %s\n", node.Address)
+	}
+	return out, nil
+}
 
 func (p *dockerProvisioner) Initialize() error {
 	err := initDockerCluster()
--- a/provision/provision.go
+++ b/provision/provision.go
@@ -240,6 +240,10 @@ type Provisioner interface {
 	RegisterUnit(Unit) error
 }
 
+type MessageProvisioner interface {
+	StartupMessage() (string, error)
+}
+
 // InitializableProvisioner is a provisioner that provides an initialization
 // method that should be called when the app is started
 type InitializableProvisioner interface {


andrewsmedina:
why not just use `tsuru-admin docker-node-list` ?

msabramo:
Good question. A few reasons.

1. A beginner doesn't know about `tsuru-admin` yet when they are first installing tsuru. I remember that when I first setup tsuru, I was following the steps to install tsuru api, gandalf, archive-server, mongodb, docker registry, etc. and I had the distinct feeling of not understanding what the purpose of all these things was and how they talked to each other. In fact, I still don't understand a lot of things like archiver-server and tsuru-unit-agent. Logging a bunch of info when the server starts up -- I feel like that might help to clarify a bit how the system works -- i.e.: the user gets some clue of what `tsr api` does, after they've installed it. I don't think I knew when I first walked through the install.

2. When the system is first set up or if things get borked, the user may not an admin account to login with to use `tsuru-admin`. Or maybe they haven't installed `tsuru-admin` yet.

3. Some folks are really into logging everything for monitoring, diagnosing, compliance, etc. At SurveyMonkey, we are big fans of Splunk and the more you make your servers log, the more data you can search through with Splunk. This is true for other stuff like logstash, kibana, etc.

cezarsa:
I'm okay with the idea of logging this information when starting the API. However, exposing cluster.Cluster in the Provisioner interface is not a good idea. We had other provisioners in the past, and we want to keep the Provisioner interface without docker specific information.

A better option would be having something like:
```
type MessageProvisioner interface {
	StartupMessage() string
}
```

And then implementing StartupMessage in dockerProvisioner to return a string containing node information. Could you make it that way?

msabramo:
@cezarsa: Updated. What do you think of it now?

api/server.go
```diff
@@ -230,6 +231,9 @@ func RunServer(dry bool) http.Handler {
 				fatal(err)
 			}
 		}
+		if startupMessageProvisioner, ok := app.Provisioner.(provision.StartupMessageProvisioner); ok {
+			startupMessageProvisioner.FprintStartupMessage(os.Stdout)
```
cezarsa: Please change this to simply return a string. The caller then can use `fmt.Printf` for now. If we start using some logging library in the future this would cause less changes.

provision/provision.go
```diff
@@ -240,6 +240,10 @@ type Provisioner interface {
 	RegisterUnit(Unit) error
 }
 
+type StartupMessageProvisioner interface {
```
cezarsa: Could you rename it to `MessageProvisioner`? It's more generic and allow us to add other messages in the future.

provision/provision.go
```diff
@@ -240,6 +240,10 @@ type Provisioner interface {
...
+	FprintStartupMessage(w io.Writer) error
```
cezarsa: Could you change this to simply `StartupMessage() (string, error)` ?

api/server.go
msabramo: Done

provision/provision.go
msabramo: Done

provision/provision.go
msabramo: Done

msabramo:
@cezarsa: Thanks for the comments! I updated:

```
$ git push -f msabramo show_docker_nodes_at_startup
Counting objects: 8, done.
Delta compression using up to 8 threads.
Compressing objects: 100% (8/8), done.
Writing objects: 100% (8/8), 1.24 KiB | 0 bytes/s, done.
Total 8 (delta 6), reused 0 (delta 0)
To git@github.com:msabramo/tsuru.git
 + b28ab9a...fa81073 show_docker_nodes_at_startup -> show_docker_nodes_at_startup (forced update)
```

msabramo:
Hmmm, don't know why Travis CI test is failing.

provision/docker/provisioner.go
```diff
@@ -40,13 +41,28 @@ func getRouterForApp(app provision.App) (router.Router, error) {
 	return router.Get(routerName)
 }
 
-type dockerProvisioner struct{}
+type dockerProvisioner struct {
+	cluster *cluster.Cluster
+}
+
+func (p *dockerProvisioner) StartupMessage() (string, error) {
+	nodeList, err := p.cluster.UnfilteredNodes()
```
cezarsa: One last change, I promise :)
We don't need `p.cluster` here. You can directly call `dockerCluster().UnfilteredNodes()` as `dockerCluster()` already takes care of initializing and using a single cluster instance for the whole package.
With this change you can remove `cluster` attribute from the struct.

msabramo: Done. Don't worry about asking for more changes if you see 'em. :smile:

msabramo:
OK, @cezarsa, Travis test is passing now, I think thanks to https://github.com/tsuru/tsuru/commit/54c4480b685f84597628ace335ccba44ecb817a5 (Thanks, @tarsisazevedo!)

cezarsa:
Thank you!

PR #1074, opened 2015-01-10
"Unknown scheme" => "Unknown auth scheme"

msabramo:
Slightly clearer since "scheme" to a lot of people implies a URL.


andrewsmedina:
@msabramo thank you!

PR #1075, opened 2015-01-10
Tweak password reset messages

msabramo:
Mostly using "reset" in place of "redefine", which looks better to this
native English speaker's eyes. Also "successfully" was misspelled before
with only one "l".


andrewsmedina:
@msabramo thank you!

PR #1076, opened 2015-01-10
Add vim modelines to tsuru conf files

msabramo:
So that vim knows that they're yaml files


msabramo:
Ping

fsouza:
Hey @msabramo, thanks for contributing! I don't think we should have a comment that is specific to an editor in our files.

Users that care about file type detection in their preferred editor, can add special comments to their own ``tsuru.conf`` file, but we should get specific here. If we start adding stuff that are specific to editors, at some point we will have to draw a line, so I'd prefer to draw the line today.

Thank you again, and sorry for not merging this one!

PR #1077, opened 2015-01-12
Allow healthcheck failures

sroze:
This PR introduce an `allowed_failures` parameter for applications' healthcheck. This is the number of allowed failures before that the health check consider the application as unhealthy (the default value is for sure 0)

This is especially interesting when sometimes there's some need of asynchronous process to be started before the application can have its normal behavior.

### DIFF
--- a/provision/docker/healthcheck.go
+++ b/provision/docker/healthcheck.go
@@ -57,7 +57,6 @@ func runHealthcheck(cont *container, w io.Writer) error {
 		status = val
 	case float64:
 		status = int(val)
-	default:
 	}
 	match, _ := hc["match"].(string)
 	if status == 0 && match == "" {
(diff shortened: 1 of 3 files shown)


provision/docker/healthcheck.go
```diff
@@ -71,6 +71,14 @@ func runHealthcheck(cont *container, w io.Writer) error {
 			return err
 		}
 	}
+	var allowed_failures int
```
andrewsmedina: You should use camel case instead of underscores.

provision/docker/healthcheck.go
```diff
@@ -71,6 +71,14 @@ func runHealthcheck(cont *container, w io.Writer) error {
...
+	switch val := hc["allowed_failures"].(type) {
+	case int:
+		allowed_failures = val
+	case float64:
+		allowed_failures = int(val)
+	default:
```
andrewsmedina: you can remove this line

andrewsmedina:
@sroze can you update the docs about `healthcheck` (http://docs.tsuru.io/en/master/using/tsuru.yaml.html#healthcheck) ?

provision/docker/healthcheck.go
sroze: So I also removed this default for `status`.

sroze:
@andrewsmedina I updated previous commit base on your comments and added the documentation bullet.

sroze:
@andrewsmedina I updated previous commit base on your comments and added the documentation bullet (and rebuilt this morning as previous Travis build failed to launch)

andrewsmedina:
@sroze thank you!

PR #1080, opened 2015-01-13
Make user-create request end with newline

msabramo:
Prevents stuff running together when using verbose mode. E.g.:

    $ tsuru-admin -v 2 user-create ${EMAIL}
    Password: ******
    Confirm: ******
    *************************** <Request uri="/users"> **********************************
    POST /users HTTP/1.1
    Host: 127.0.0.1:8080
    Connection: close
    Authorization: bearer d2f838d903e100b56beab35d76718924065d7a36

    {"email":"me@mycompany.com", "password":"secret"}*************************** </Request uri="/users"> **********************************
    *************************** <Response uri="/users"> **********************************
    HTTP/1.1 500 Internal Server Error
    ...

### DIFF
--- a/cmd/client.go
+++ b/cmd/client.go
@@ -60,6 +60,9 @@ func (c *Client) Do(request *http.Request) (*http.Response, error) {
 			return nil, err
 		}
 		fmt.Fprintf(c.context.Stdout, string(requestDump))
+		if requestDump[len(requestDump)-1] != '\n' {
+			fmt.Fprintln(c.context.Stdout)
+		}
 		fmt.Fprintf(c.context.Stdout, "*************************** </Request uri=%q> **********************************\n", request.URL.RequestURI())
 	}
 	response, err := c.HTTPClient.Do(request)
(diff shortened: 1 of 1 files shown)


cezarsa:
It would be better to handle this inside while handling the verbosity code. The json is still valid without the "\n" and besides these cases, we can't control how `json.Marshal()` (doesn't add \n) and `json.NewEncoder().Encode()` (adds \n) will work in the future.

So, to be safe, adding something like that while dumping the request would be better:
```
if requestDump[len(requestDump)-1] != '\n' {
	fmt.Println()
}
```

msabramo:
Updated to use this:

```diff
diff --git a/cmd/client.go b/cmd/client.go
index 75d5148..dfcf3e4 100644
--- a/cmd/client.go
+++ b/cmd/client.go
@@ -60,6 +60,9 @@ func (c *Client) Do(request *http.Request) (*http.Response, error) {
                        return nil, err
                }
                fmt.Fprintf(c.context.Stdout, string(requestDump))
+               if requestDump[len(requestDump)-1] != '\n' {
+                       fmt.Fprintln(c.context.Stdout)
+               }
                fmt.Fprintf(c.context.Stdout, "*************************** </Request uri=%q> **********************************\n", request.URL.RequestURI())
        }
        response, err := c.HTTPClient.Do(request)
```

cezarsa:
Could you change the commit message to reflect this change? Also do you think it would be useful to apply the same handling to the code that dumps the response?

msabramo:
Yep. Updated.

cezarsa:
Thanks!

PR #1081, opened 2015-01-13
Print gandalf failure in server log

msabramo:
It's nice to see this sort of error (involving inability to connect to another component) on the server side, since it may indicate a misconfiguration or a system problem.

### DIFF
--- a/auth/user.go
+++ b/auth/user.go
@@ -245,6 +245,7 @@ func (u *User) CreateOnGandalf() error {
 	gURL := repository.ServerURL()
 	c := gandalf.Client{Endpoint: gURL}
 	if _, err := c.NewUser(u.Email, keyToMap(u.Keys)); err != nil {
+		fmt.Printf("Failed to create user in the git server: %s\n", err)
 		return fmt.Errorf("Failed to create user in the git server: %s", err)
 	}
 	return nil


cezarsa:
This error is already logged using the configured logger in the method calling CreateOnGandalf(), see https://github.com/tsuru/tsuru/blob/master/api/auth.go#L77-L85

msabramo:
Ah okay, thanks. I suspect what is happening is that it's sending that message to the syslog where I didn't see it.

If I set this in `/etc/tsuru/tsuru.conf`:

```yaml
log:
    file: /dev/stdout
```

then I get a nice error on stdout:

```
2015/01/15 14:24:21 ERROR: error trying to create user "user@company.com" in gandalf (http://127.0.0.1:8000): Failed to create user in the git server: Failed to connect to Gandalf server (http://127.0.0.1:8000) - Post http://127.0.0.1:8000/user: dial tcp 127.0.0.1:8000: connection refused
2015/01/15 14:24:21 ERROR: Failed to create user in the git server: Failed to connect to Gandalf server (http://127.0.0.1:8000) - Post http://127.0.0.1:8000/user: dial tcp 127.0.0.1:8000: connection refused
```

I wonder if:

```yaml
log:
    file: /dev/stdout
```

should be the default since maybe folks are more likely to notice an error on stdout than in the syslog?

PR #1087, opened 2015-01-15
.travis.yml: Generalize sed so works with forks

msabramo:
My attempt at making Travis not error out on my tsuru fork.


andrewsmedina:
@msabramo thank you!

Issue #1111, opened 2015-02-26
app: support removing all units from the application


fsouza:
Sometimes it may be useful to have 0 units.

lucasweiblen:
does this means just to remove the following lines?

https://github.com/tsuru/tsuru/blob/master/app/app.go#L402
https://github.com/tsuru/tsuru/blob/master/app/app.go#L403
https://github.com/tsuru/tsuru/blob/master/app/app.go#L404

fsouza:
I think so, running some tests can prove it

FTR, you can link a selected area: https://github.com/tsuru/tsuru/blob/master/app/app.go#L402:L404.

lucasweiblen:
heheh didnt know how to select a area :+1: 

I would like to try to solve this issue but I am having some issues to run all tests inside the box provisioned by tsuru-bootstrap.

Do you have any idea what could possible be happening?
I tried go get ./... and make test , both returning some issues.

fsouza:
What's the output you get when running ``go test ./...``?

lucasweiblen:
browsing the issues I found #1018 .. if I remember well I'm having the same problems but almost sure Mongo and Redis were running.. will try again tonight!

lucasweiblen:
well, it seems Mongo and Redis were not running when I tried yesterday hehe, now I'm managing to run the tests.. thanks

think I will handle #1018 as well

fsouza:
Closed via #1118.

PR #1118, opened 2015-03-04
app: support removing all units from the application #1111

lucasweiblen:
guys, could you please review this PR? if there is anything missing please say me. I'm just getting used to tsuru's codebase yet.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -398,9 +398,9 @@ func (app *App) RemoveUnits(n uint) error {
 	if n == 0 {
 		ReleaseApplicationLock(app.Name)
 		return stderr.New("Cannot remove zero units.")
-	} else if l := uint(len(app.Units())); n > l {
+	} else if length := uint(len(app.Units())); n > length {
 		ReleaseApplicationLock(app.Name)
-		return fmt.Errorf("Cannot remove %d units from this app, it has only %d units.", n, l)
+		return fmt.Errorf("Cannot remove %d units from this app, it has only %d units.", n, length)
 	}
 	go func() {
 		defer ReleaseApplicationLock(app.Name)
--- a/app/app.go
+++ b/app/app.go
@@ -398,10 +398,7 @@ func (app *App) RemoveUnits(n uint) error {
 	if n == 0 {
 		ReleaseApplicationLock(app.Name)
 		return stderr.New("Cannot remove zero units.")
-	} else if l := uint(len(app.Units())); l == n {
-		ReleaseApplicationLock(app.Name)
-		return stderr.New("Cannot remove all units from an app.")
-	} else if n > l {
+	} else if l := uint(len(app.Units())); n > l {
 		ReleaseApplicationLock(app.Name)
 		return fmt.Errorf("Cannot remove %d units from this app, it has only %d units.", n, l)
 	}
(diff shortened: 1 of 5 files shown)


andrewsmedina:
LGTM.

app/app.go
```diff
@@ -398,10 +398,7 @@ func (app *App) RemoveUnits(n uint) error {
 	if n == 0 {
 		ReleaseApplicationLock(app.Name)
 		return stderr.New("Cannot remove zero units.")
-	} else if l := uint(len(app.Units())); l == n {
-		ReleaseApplicationLock(app.Name)
-		return stderr.New("Cannot remove all units from an app.")
-	} else if n > l {
+	} else if l := uint(len(app.Units())); n > l {
```
fsouza: The change looks good, but I'd like to take this opportunity to rename this variable to "length". Using ``l`` as a variable name is confuse :-(

And I'm sorry, because I think I wrote this code in the first place. Do you mind just renaming this variable? After that, we can merge the pull request.

cezarsa:
Also, please notice that this PR solves half of the problem. After removing all units from an application, one should be able to call `tsuru unit-add` to bring the application back to life.
However, there's a check in the provisioner which doesn't allow adding units to applications with 0 units. This could be handled in another PR, but it has to go before everything works.

app/app.go
lucasweiblen: no problems, man.. will fix this tonight

lucasweiblen:
hey @cezarsa, thanks for the feedback!
I will try to find this check in the provisioner in order to handle this second half of the problem.

lucasweiblen:
@cezarsa could you please confirm if this is the check you were talking about? thanks

https://github.com/tsuru/tsuru/blob/master/provision/docker/provisioner.go#L564:L566

cezarsa:
@lucasweiblen Yes, this check has to be changed. We cannot simply erase it because we still need to make sure the application have at least 1 deploy before adding new units. So we could change `length < 1` to `a.GetDeploys() == 0`. This might have a bit of an impact in this package's tests.

lucasweiblen:
thanks for the tip @cezarsa, I will have a look at this tonight.

lucasweiblen:
@cezarsa as you said , some tests(4 or 5) are now failing because the number of deploys is 0. To setup a deploy in the tests I would need to do something like this, right?

https://github.com/tsuru/tsuru/blob/master/provision/docker/provisioner_test.go#L104:L129

Is there a better way to do that?

cezarsa:
@lucasweiblen I think a more simple alternative is making the deploys attribute in FakeApp public, here:
https://github.com/tsuru/tsuru/blob/master/provision/provisiontest/fake_provisioner.go#L39
And in tests calling `AddUnits` test do something like:
```
app := provisiontest.NewFakeApp(...)
app.Deploys = 1
```

lucasweiblen:
@cezarsa much simpler approach! thank you very much.

I have changed AddUnits and fixed the broken tests. If anything is still missing please let me know!

fsouza:
@lucasweiblen the code looks good, can you rename the test "TestProvisionerAddUnitsWithoutContainers" to "TestProvisionerAddUnitsWithNoDeploys"?

Thanks!

lucasweiblen:
@fsouza done

fsouza:
LGTM.

Let's wait for @cezarsa :)

cezarsa:
Thanks a lot Lucas.

Issue #1119, opened 2015-03-04
app-list returns JSON error while app removal is running in background


fsouza:
If I run app-remove and immediately after run app-list, I get the error "Error: json: error calling MarshalJSON for type app.App: repository not found":

```
% tsuru app-remove -ya xikin
App "xikin" successfully removed!
% tsuru app-list
Error: json: error calling MarshalJSON for type app.App: repository not found
% tsuru app-info -a xikin
Error: json: error calling MarshalJSON for type *app.App: repository not found
# wait a little bit
% tsuru app-info -a xikin
Error: App xikin not found.
% tsuru app-list
+-----------------------------------+---------------------------+-------------------------------------------------------------+--------+
| Application                       | Units State Summary       | Address                                                     | Ready? |
+-----------------------------------+---------------------------+-------------------------------------------------------------+--------+
| mysuperapp                        | 1 of 1 units in-service   | mysuperapp.tsuru.mycompany.com                              | Yes    |
+-----------------------------------+---------------------------+-------------------------------------------------------------+--------+
```

It also fails for app-info. IMO, it's ok to ignore "repository not found" failure in the app MarshalJSON method, so I'm marking this as "easy".

andrewsmedina:
closed by #1121

PR #1121, opened 2015-03-05
app-list returns JSON error while app removal is running in background #1119

lucasweiblen:
@fsouza is this what you suggested?

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -90,10 +90,7 @@ func (app *App) Units() []provision.Unit {
 
 // MarshalJSON marshals the app in json format.
 func (app *App) MarshalJSON() ([]byte, error) {
-	repo, err := repository.Manager().GetRepository(app.Name)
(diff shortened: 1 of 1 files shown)


andrewsmedina:
@lucasweiblen thank you!

fsouza:
@lucasweiblen could you please add a test for this? This test can help us in the case we change ``GetRepository`` to return a pointer, for example (currently it doesn't, so this code is still safe).

You can do something like [TestAppMarshalJSON](https://github.com/tsuru/tsuru/blob/a4f767220bdd21de70e90caac032af438e29fb63/app/app_test.go#L1877:L1937), but without the call to ``repository.Manager().CreateRepository("name", nil)``, and expecting the repository to be empty.

Thanks for contributing!

lucasweiblen:
@fsouza sure... I will add the test.

PR #1122, opened 2015-03-05
Change `tsuru-admin` commands by `tsuru` after installation

sroze:
The documentation is wrong since the last release, this PR updates it.


andrewsmedina:
@sroze thank you!

Issue #1125, opened 2015-03-08
improve error message for MongoDB connection handling


fsouza:
Issue #1124 was easy to fix, but because we didn't provide a good error message, the user had a hard time trying to fix it.

We should report a better error than "no reachable servers". Even a "mongodb" prefix would make it easier to know that the issue is with MongoDB.

So, a message like "mongodb: no reachable servers" should work, and is quite easy to do.

fsouza:
Closed via #1126.

PR #1126, opened 2015-03-08
db/storage: improve error message for mongodb connection handling

lucasweiblen:
#1125

### DIFF
--- a/db/storage/storage.go
+++ b/db/storage/storage.go
@@ -7,6 +7,7 @@ package storage
 import (
 	"sync"
 	"time"
+	"errors"
 
 	"gopkg.in/mgo.v2"
 )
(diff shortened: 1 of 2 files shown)


db/storage/storage.go
```diff
@@ -45,7 +46,7 @@ func (c *Collection) Close() {
 func open(addr, dbname string) (*Storage, error) {
 	sess, err := mgo.Dial(addr)
 	if err != nil {
-		return nil, err
+		return nil, errors.New("mongodb: no reachable servers")
```
fsouza: I'd rather do ```fmt.Errorf("mongodb: %s", err)```, so we can catch other failures in mgo.Dial as well (including authentication errors).

lucasweiblen: yep! :smile:

andrewsmedina:
@lucasweiblen thank you!

Issue #1113, opened 2015-02-27
cmd/tsuru-admin: add command to list routers


fsouza:
We should have a command for listing the available routers for creating new plans:

```
% tsuru-admin router-list
```

lucasweiblen:
anybody working on this issue?

fsouza:
Nope, you can work on it, if you will.

msabramo:
@lucasweiblen is shredding it up lately! Thanks for all your work!

lucasweiblen:
I will give it a try!
Thanks @msabramo :)

dcarley:
Lucas completed this in tsuru/tsuru-admin#2

fsouza:
@dcarley thank you for pointing it!

PR #1139, opened 2015-03-14
#1113 - add command to list routers

lucasweiblen:
Guys, could you please review this?

I've added the handler and the function to list routers but I'm not sure if they are in the right place.
I have also a doubt related to the routers properties on the config file. Is "type" the only property?

I've already done a draft of the code to add the command on tsuru-admin, as soon as I fix/improve this PR I will open the pull request there.

### DIFF
--- a/api/plan.go
+++ b/api/plan.go
@@ -61,3 +61,12 @@ func removePlan(w http.ResponseWriter, r *http.Request, t auth.Token) error {
 	}
 	return err
 }
+
+func listRouters(w http.ResponseWriter, r *http.Request, t auth.Token) error {
+	routers, err := app.RoutersList()
+	if err != nil {
+		return err
+	}
+	w.Header().Set("Content-Type", "application/json")
+	return json.NewEncoder(w).Encode(routers)
+}
--- a/api/plan.go
+++ b/api/plan.go
@@ -64,7 +64,7 @@ func removePlan(w http.ResponseWriter, r *http.Request, t auth.Token) error {
 }
 
 func listRouters(w http.ResponseWriter, r *http.Request, t auth.Token) error {
-	routers, err := router.GetList()
+	routers, err := router.List()
 	if err != nil {
 		return err
 	}
--- a/api/plan.go
+++ b/api/plan.go
@@ -1,4 +1,4 @@
-// Copyright 2014 tsuru authors. All rights reserved.
+// Copyright 2015 tsuru authors. All rights reserved.
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
@@ -11,6 +11,7 @@ import (
 	"github.com/tsuru/tsuru/app"
 	"github.com/tsuru/tsuru/auth"
 	"github.com/tsuru/tsuru/errors"
+	"github.com/tsuru/tsuru/router"
 )
 
 func addPlan(w http.ResponseWriter, r *http.Request, t auth.Token) error {
@@ -63,7 +64,7 @@ func removePlan(w http.ResponseWriter, r *http.Request, t auth.Token) error {
 }
 
 func listRouters(w http.ResponseWriter, r *http.Request, t auth.Token) error {
-	routers, err := app.RoutersList()
+	routers, err := router.GetList()
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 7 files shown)


app/plan.go
```diff
@@ -144,3 +145,28 @@ func PlanRemove(planName string) error {
 	}
 	return err
 }
+
+type Router struct {
+	Name string `json:"name"`
+	Type string `json:"type"`
+}
+
+func RoutersList() ([]Router, error) {
```
fsouza: I don't think that this function belongs to this package. It should be in the package "router".

app/plan.go
```diff
@@ -144,3 +145,28 @@ func PlanRemove(planName string) error {
...
+	routerConfig, err := config.Get("routers")
+	if err != nil {
+		return nil, err
+	}
+	routers, _ := routerConfig.(map[interface{}]interface{})
+	var routersList []Router
```
fsouza: You can prealloc this list with:

```
routersList := make([]Router, 0, len(routers))
```

app/plan.go
lucasweiblen: thanks, I will move to the right package :smile:

api/plan.go
```diff
@@ -61,3 +61,12 @@ func removePlan(w http.ResponseWriter, r *http.Request, t auth.Token) error {
 	}
 	return err
 }
+
+func listRouters(w http.ResponseWriter, r *http.Request, t auth.Token) error {
```
andrewsmedina: you can put this handler in a `router.go`.

lucasweiblen: thanks @andrewsmedina 
there is already a router.go file in the api package but it's a file used for registering the handlers with Gorilla Mux router (or some stuff like that ..). 

https://github.com/tsuru/tsuru/blob/master/api/router.go

If I put there it will be strange, no? Is it ok if I rename this existent router.go(with delayedRouter stuff) file to delayed_router.go and just use the router.go file for the listRouters handler?

andrewsmedina: I don't know. @fsouza what do you think about it?

cezarsa: I think you can keep it there. router.go is completely unrelated in this package. It makes a certain sense to keep it in plan.go since routers are part of plans.

api/plan_test.go
```diff
@@ -171,3 +172,26 @@ func (s *S) TestPlanRemoveInvalid(c *check.C) {
 	m.ServeHTTP(recorder, request)
 	c.Assert(recorder.Code, check.Equals, http.StatusNotFound)
 }
+
+func (s *S) TestRoutersList(c *check.C) {
```
andrewsmedina: Remember to add a test to check that this handler is accessed only by admins.

api/plan.go
lucasweiblen: :+1: @cezarsa

api/plan_test.go
lucasweiblen: sure!

router/router.go
```diff
@@ -193,3 +194,28 @@ func Swap(r Router, backend1, backend2 string) error {
 	}
 	return swapBackendName(backend1, backend2)
 }
+
```
lucasweiblen: had to change the struct name to PlanRouter because of the interface Router. is there a better way to handle this?

router/router.go
```diff
@@ -193,3 +194,28 @@ func Swap(r Router, backend1, backend2 string) error {
...
+type PlanRouter struct {
```
lucasweiblen: had to change the struct name to PlanRouter because of the interface Router. is there a better way to handle this?

router/router.go
```diff
@@ -193,3 +194,28 @@ func Swap(r Router, backend1, backend2 string) error {
...
+	Name string `json:"name"`
+	Type string `json:"type"`
+}
+
+func GetList() ([]PlanRouter, error) {
```
fsouza: I think that just "List" is better.

fsouza:
LGTM

@cezarsa @andrewsmedina any objections to merging it?

andrewsmedina:
@fsouza LGTM

cezarsa:
Thanks Lucas!

Issue #1134, opened 2015-03-12
cmd/tsuru: add filters in app-list


andrewsmedina:
`app-list` list all application that the logged user has access. It would be nice if `app-list` had some filters. `grep` solves this problem when you need to filter the apps by name, but sometimes you will need to list all apps that uses a specific platform or by teamowner, eg:

    $ tsuru app-list --platform=python

We can start adding `platform`, `teamowner`, `owner`, `name` filters. Where the name can accept a regex pattern.

lucasweiblen:
anyone working on this issue?

andrewsmedina:
@lucasweiblen no. Feel free to work on in this issue.

lucasweiblen:
nice

PR #1143, opened 2015-03-22
cmd/tsuru: add filters in app-list 

lucasweiblen:
#1134

### DIFF
--- a/api/suite_test.go
+++ b/api/suite_test.go
@@ -97,8 +97,6 @@ func (s *S) SetUpSuite(c *check.C) {
 	app.AuthScheme = nativeScheme
 	p := app.Platform{Name: "zend"}
 	s.conn.Platforms().Insert(p)
-	p2 := app.Platform{Name: "python"}
-	s.conn.Platforms().Insert(p2)
 }
 
 func (s *S) TearDownSuite(c *check.C) {
--- a/api/suite_test.go
+++ b/api/suite_test.go
@@ -97,6 +97,8 @@ func (s *S) SetUpSuite(c *check.C) {
 	app.AuthScheme = nativeScheme
 	p := app.Platform{Name: "zend"}
 	s.conn.Platforms().Insert(p)
+	p2 := app.Platform{Name: "python"}
+	s.conn.Platforms().Insert(p2)
 }
 
 func (s *S) TearDownSuite(c *check.C) {
@@ -106,6 +108,7 @@ func (s *S) TearDownSuite(c *check.C) {
 func (s *S) SetUpTest(c *check.C) {
 	repositorytest.Reset()
 	repository.Manager().CreateUser(s.user.Email)
+	repository.Manager().CreateUser(s.adminuser.Email)
 	factory, err := queue.Factory()
 	c.Assert(err, check.IsNil)
 	fakeFactory := factory.(*queuetest.FakePubSubQFactory)
--- a/app/app.go
+++ b/app/app.go
@@ -1038,19 +1038,26 @@ func (app *App) LastLogs(lines int, filterLog Applog) ([]Applog, error) {
 	return logs, nil
 }
 
-func buildAppListFilterQuery(filterApp *App) bson.M {
+type Filter struct {
+	Name      *regexp.Regexp
+	Platform  string
+	TeamOwner string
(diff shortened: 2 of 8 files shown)


fsouza:
Hi @lucasweiblen, thanks for working on this.

I'd prefer to have a "Filter" type, that declares the filters explicitly:

```
type Filter struct {
    Name      *regexp.Regexp
    Platform  string
    TeamOwner string
    UserOwner string
}
```

And then declare a method in the filter that builds the query:

```
func (f *Filter) Query() bson.M {
	return bson.M{
		"name":      bson.M{"$regexp": f.Name.String()},
		"platform":  f.Platform,
		"teamowner": f.TeamOwner,
		"owner":     f.UserOwner,
	}
}
```

I know it doesn't look like it makes much difference, but I think it's a better choice when talking about semantics because (1) it's explicit that the name is a regexp (2) the type names makes it explicit that it's is a filter.

What do you think?

lucasweiblen:
yeah, I agree with you @fsouza! Thanks for the feedback and suggestions.

app/app.go
```diff
@@ -1038,19 +1038,51 @@ func (app *App) LastLogs(lines int, filterLog Applog) ([]Applog, error) {
 	return logs, nil
 }
 
+type Filter struct {
+	Name      *regexp.Regexp
+	Platform  string
+	TeamOwner string
+	UserOwner string
+}
+
```
lucasweiblen: @fsouza tried here your suggestion for the query method which was way cleaner but somehow api tests were breaking  because the query was returning 0 elements.. don't know all corners of mgo so perhaps I'm missing something here..

api/suite_test.go
```diff
@@ -97,6 +97,8 @@ func (s *S) SetUpSuite(c *check.C) {
 	app.AuthScheme = nativeScheme
 	p := app.Platform{Name: "zend"}
 	s.conn.Platforms().Insert(p)
```
lucasweiblen: add this to test filtering of platforms in api_test. If there's a better way to do this let me know.

app/app.go
andrewsmedina: @lucasweiblen all api tests pass here.

lucasweiblen: hi @andrewsmedina .. yes, the solution that I've found is working, the tests are passing but its not as clean as the solution @fsouza suggested:

```
func (f *Filter) Query() bson.M {
    return bson.M{
        "name":      bson.M{"$regexp": f.Name.String()},
        "platform":  f.Platform,
        "teamowner": f.TeamOwner,
        "owner":     f.UserOwner,
    }
}
```

So if you have any suggestions to improve my solution , please let me know :)

app/app.go
```diff
@@ -1038,19 +1038,51 @@ func (app *App) LastLogs(lines int, filterLog Applog) ([]Applog, error) {
 	return logs, nil
 }
 
+type Filter struct {
+	Name      *regexp.Regexp
```
cezarsa: It's safer for this field to be a string. Go regexp engine isn't completely compatible with mongodb engine, so a valid mongodb regexp might not be compilable by Go. Turning this into a string will let MongoDB be responsible for parsing and validating the regexp.

lucasweiblen: Good point @cezarsa. Thanks! I will change this.

app/app.go
fsouza: Hi @lucasweiblen, this solution wouldn't actually work, because that would ask MongoDB to give apps that have platform == "", the solution you proposed is the right one.

api/suite_test.go
fsouza: That's fine, if you prefer, you can add the new platform only in the test that you're going to use it.

lucasweiblen: cool! I'll move this to the specific test, it makes more sense to add the platform just there.

fsouza: Please ensure you have a deferred call for removing the platform. Thank you!

fsouza:
Thank you!

fsouza:
Are you going to send a pull request to the tsuru-client repository as well?

lucasweiblen:
yes, I'm planning to start working there today

fsouza:
Thank you!

Issue #1145, opened 2015-03-25
tsuru-admin: docker-pool-teams-add returns bad error when pool doesn't exist


fsouza:
```
% tsuru-admin docker-pool-teams-add pool-that-doesnt-exist some-team
Error: not found
```

It's hard to tell what doesn't exist. We know that it is the pool because it doesn't check if the team exists, but we could improve the message with something like "Error: pool not found".

fsouza:
Fixed by #1155.

PR #1155, opened 2015-03-29
tsuru-admin: docker-pool-teams-add returns bad error when pool doesn't exist

lucasweiblen:
#1145


andrewsmedina:
@lucasweiblen thank you!

fsouza:
Thank you! Please include something like "Closes #[issue-number]" in future commit messages, so GitHub will automatically close the issue when the pull request gets merged!

You can also use other terms like "Fix/Fixes" (which is better for issues that are bugs, not the case for #1145).

lucasweiblen:
nice, didn't know that! thanks

Issue #1152, opened 2015-03-26
cmd/tsuru: unify service-info and service-doc


fsouza:
``tsuru service-info`` could display the documentation of the service, if any is available.

lucasweiblen:
in this case we would need to display documentation for the service AND service-info right? Is there a better way to do this than returning the JSON representation of a map? 
something like: 
`
b, _ := json.Marshal(instances)
serviceMap := map[string][]byte{"doc": s.Doc, "info": b}
b2, _ := json.Marshal(serviceMap)
w.Write(b2)
return nil
`

andrewsmedina:
You can send the serialized "Service" and add a "Instances" attribute.

lucasweiblen:
good idea! thanx @andrewsmedina

PR #1174, opened 2015-04-18
cmd/tsuru: unify service-info and service-doc

lucasweiblen:
#1152 
that's it for the tsuru part of this issue, right? If there's something missing or wrong please point me in the right direction.
I will open a PR for the tsuru-client part as well in the next days.

### DIFF
--- a/service/service.go
+++ b/service/service.go
@@ -22,6 +22,7 @@ type Service struct {
 	Teams        []string
 	Doc          string
 	IsRestricted bool `bson:"is_restricted"`
+	Instances    []ServiceInstance
 }
 
 var (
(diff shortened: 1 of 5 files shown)


service/service.go
```diff
@@ -22,6 +22,7 @@ type Service struct {
 	Teams        []string
 	Doc          string
 	IsRestricted bool `bson:"is_restricted"`
+	Instances    []ServiceInstance
```
andrewsmedina: I'm -1 for to add this field.

I believe that the better approach is to add a `service.FindServiceInstanceByName("name", &user)` method and to split this handler in two handlers, one to get the service info and another to get service instances by service name.

lucasweiblen: cool, Andrews. I agree it's a better approach. :smile: 
Let's see if the guys have any other suggestions and I can start working on improving this PR.

Thanks for your feedback.

fsouza: I don't think there's much else to review. With Andrews' suggestion, most changes move to the client, that will now need to issue to API calls and combine the results.

So, you will:

1. remove the service doc handler (or convert it into a handler that returns full information of the service)
1. add a handler for listing instances of a service (that the user has access to, obviously)

And then comes the client changes.

Thank you!

lucasweiblen: thanks @fsouza.. will work on these improvements :smile:

fsouza:
I will close this pull request, as @raulgbrmf is currently working on an alternative solution.

@lucasweiblen thanks for contributing!

PR #1179, opened 2015-04-27
cmd: increase the chance of a command being suggested

scorphus:
What do you think of, instead of just the prefix, consider any occurrence of the argument in the command name when deciding to suggest the command or not. Here is a brief example:

* Before this change:
```
► tsuru log
tsuru: "log" is not a tsuru command. See "tsuru help".

Did you mean?
    login
    logout
```

* After this change:
```
► tsuru log
tsuru: "log" is not a tsuru command. See "tsuru help".

Did you mean?
    app-log
    login
    logout
```

As far as I could test, [tsuru/tsuru-client](http://github.com/tsuru/tsuru-client) wasn't affected by this change. Please advise me if any other project is, though.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -154,7 +154,7 @@ func (m *Manager) Run(args []string) {
 					sort.Strings(keys)
 					for _, key := range keys {
 						levenshtein := fuzzy.Levenshtein(&key, &args[0])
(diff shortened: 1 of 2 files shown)


andrewsmedina:
@scorphus wow! thank you!

scorphus:
YAY! My pleasure! :wink:

PR #1180, opened 2015-04-27
Correct small typo in Gandalf set-up documentation

annashipman:
There are three steps not two.

### DIFF
--- a/docs/installing/gandalf.rst
+++ b/docs/installing/gandalf.rst
@@ -100,8 +100,7 @@ Then start gandalf and archive-server:
 Configuring tsuru to use Gandalf
 ================================
 
-In order to use Gandalf, you need to change tsuru.conf accordingly, it's a
-two steps setup:
+In order to use Gandalf, you need to change tsuru.conf accordingly:
 
 #. Define "repo-manager" to use "gandalf";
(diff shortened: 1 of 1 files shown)


fsouza:
Thank you!

fsouza:
I think that all of this "it's a ..." is not necessary. We could go with just "In order to use Gandalf, you need to change tsuru.conf accordingly:".

This way we don't need to change the number again if we add or remove a bullet.

What do you think?

annashipman:
Oh yes, good thinking. I'll rebase.

I can't see why Travis has failed.

annashipman:
I've updated it as you suggested.

fsouza:
The failure is probably from an intermittent test:

```
----------------------------------------------------------------------
FAIL: auto_scale_test.go:334: S.TestAutoScaleConfigRunRebalanceOnly

auto_scale_test.go:403:
    c.Assert(evts, check.HasLen, 1)
... obtained []docker.autoScaleEvent = []docker.autoScaleEvent(nil)
... n int = 1
OOPS: 331 passed, 1 FAILED
--- FAIL: Test (41.13s)
```

@cezarsa have you seen this failure before?

fsouza:
@annashipman merging it, thank you!

PR #1214, opened 2015-06-17
Error when used the suffix domain

leandrosouza:
This error occurs when you try to use the suffix domain like cname.domain.com to create cname.domain.com.br

another example:
cname.domain.com
cnamenew.domain.com


andrewsmedina:
@leandrosouza can you add a test for this case?

PR #1219, opened 2015-06-24
Fixed scope parsing.

RichardKnop:
I assume the scope in config would look something like this:

```yaml
auth:
  oauth2:
    scope: user-create team-user-add
```

Since config.GetString("auth:oauth:scope") is a string, when passing it to oauth2.Config.Scopes, it should be split by spaces rather then passed as the only slice element.

See the Oauth2 spec: https://tools.ietf.org/html/rfc6749#page-23

Scope is a space delimited string describing permissions.

What do you think? Does this make sense?

Later down the road, I believe Oauth2's scope could be used for giving users more granular permissions.

### DIFF
--- a/auth/oauth/oauth.go
+++ b/auth/oauth/oauth.go
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"net/http"
 	"strconv"
+	"strings"
 
 	"github.com/tsuru/config"
 	"github.com/tsuru/tsuru/auth"
@@ -83,7 +84,7 @@ func (s *OAuthScheme) loadConfig() (oauth2.Config, error) {
 	s.BaseConfig = oauth2.Config{
 		ClientID:     clientId,
 		ClientSecret: clientSecret,
-		Scopes:       []string{scope},
+		Scopes:       strings.Split(scope, " "),
 		Endpoint: oauth2.Endpoint{
 			AuthURL:  authURL,
 			TokenURL: tokenURL,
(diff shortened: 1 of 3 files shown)


cezarsa:
Hi @RichardKnop thanks for contributing. I think it makes sense.
Could you only add a test case for this behavior? It should be similar to TestOAuthLogin in oauth_test.go only setting the scope config to multiple values before running.

RichardKnop:
@cezarsa How about now?

I have added a unit test. Modified the code a little bit.

cezarsa:
There's no need to change all this because of tests. The previous version was fine.
In tests, instead of:
```
params["scope"] = "user-create team-user-create"
```
simply call:
```
config.Set("auth:oauth:scope", "user-create team-user-create")
defer config.Set("auth:oauth:scope", "myscope")
```
The `defer` is just to set the value back to what was set in SetUpSuite and prevent other tests from being affected. 
This way you won't have to change the Token struct neither handleToken signature.
In the `c.Assert(s.bodies[0],...` assert you'll make sure the proper scope was send on the request.

RichardKnop:
Yes the problem is, if I don't change the Token struct, the scope is not persisted in the database. It's not a public property.

cezarsa:
Why would you need the scope to be stored in the database? It will already be available in the config file. If it's value is changed in the config file users won't be able to authenticate and will have to log in again.

RichardKnop:
However in Oauth2 you can request a different scope when requesting access token. You can pass a scope in query string or POST body, otherwise it will fall back to the default scope.

RichardKnop:
@cezarsa I'm closing this for now. I might need to rethink how to approach this.

Issue #1220, opened 2015-06-24
Feature request / discussion: User permissions


dhilton:
We have a use case for being able to restrict create-user permission to a new class / role of user, we call a team lead(s). This role allows us to delegate admin from a central platform team, to team leads, restricting who can add a user to the platform and thus a team. 

Has there been any discussions / thoughts about this use case before or any prior art to adding such a model to the platform? We've looked at how the admin team works and understand the model employed here - if we were to implement such a model, what would be the preferences for how this might be done?

dhilton:
Our current thinking is that option 2 gets us to a workable place quicker, but we note that option 3 is a longer term goal. We've looked at https://github.com/shavac/go.sec - specifically adding the ACL and RBAC mentioned there as it comes with a MongoDB engine, which makes integrating it with the Server application in Tsuru easier. This option 3 however, will require quite a change we reckon based on an initial view formed by reading the code of the client, admin client and the dashboard. 

Cloud Foundry for example implements a similar organisation, space, team and user model http://docs.pivotal.io/pivotalcf/concepts/roles.html 

The scenarios we see :

* TL installs the Tsuru client
* TL creates Users on Tsuru using the Tsuru client - one for each member of the team
* TL adds the new users to the correct Team
* TL will manually send login details to Users
* TL can remove Users if someone leaves the Team
* TL will need to monitor their own Quota this is not part of the scope of this story
* Normal Devs cannot allocate a Team to a Pool
* TL cannot allocate a Team to a Pool
* TL can call team-user-remove
* Normal Dev cannot call team-user-remove
* Normal Dev cannot add new Users
* Normal Dev cannot add Users to a Team
* Normal Dev cannot app-set-team-owner to change the team owner
* TL would like to unlock apps if they are locked - this is not in current scope

Where TL is a "Team lead" role. A team can have multiple team leads who are a subset of the team members. 

This type of change would allow us to operate a federated, self-service model across multiple parts of our organisation.

andrewsmedina:
Thinking more about this issue I believe that we can solve it in a flexible way introducing roles and permissions.

An user can have more than one role, and a role can have one or more permissions. In this way "team leader" can be a "role".

What do you think about it?

dhilton:
[...]

PR #1222, opened 2015-06-26
Adding Concept Of Team Leads

RichardKnop:
See this issue: https://github.com/tsuru/tsuru/issues/1220

This pull request adds a new slice on Team struct which can contain list of team leads. A new API endpoint has been introduced to add and remove team leads from a team.

- each team can have multiple team leads.
- team leads can be added or removed by admin or another team lead from the same team
- removing last team lead will return error
- removing a user now also removes the user from team leads

This is a first step towards more granular team permissions.

Tell us what you think and if you have any objections.

In the future, we would follow up with more PRs to continue developing the team lead functionality.

If you accept this PR, we would then add new commands to tsuru client to add and remove team leads as well as improve team-user-list command to indicate which users are team leads.

### DIFF
--- a/api/auth.go
+++ b/api/auth.go
@@ -258,6 +258,18 @@ func addUserToTeamInRepository(user *auth.User, t *auth.Team) error {
 	return nil
 }
 
+func addLeadToTeamInDatabase(user *auth.User, team *auth.Team) error {
+	if err := team.AddTeamLead(user); err != nil {
+		return &errors.HTTP{Code: http.StatusConflict, Message: err.Error()}
+	}
+	conn, err := db.Conn()
+	if err != nil {
+		return err
+	}
+	defer conn.Close()
+	return conn.Teams().UpdateId(team.Name, team)
+}
+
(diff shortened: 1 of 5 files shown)


andrewsmedina:
hi @RichardKnop, your code looks good to me. But I think that to add team leader to an team is a not flexible way to solve the authorization problem.

My proposal is to introduce roles and permissions, that solve your problem in a flexible way. An user can have more than one role, and a role can have one or more permissions. In this way "team leader" can be a "role".

What do you think about it?

RichardKnop:
@andrewsmedina I think that would be better approach. Do you have a road map for proper roles and permissions?

We opted for a simpler approach of team leads to solve our current use case but adding proper acl would be preferable in the long run. However adding roles and permissions would be a big task.

RichardKnop:
@andrewsmedina Btw, do you know why the Travis build fails for Go tip but works on 1.4? All tests pass.

```go build -x -o tsr ./cmd/tsr```

Works on my Mac.

andrewsmedina:
@RichardKnop we will add the issue #1220 to be done in the milestone after 0.12. ok?

PR #1226, opened 2015-07-03
router/vulcand: Test against memng instead of etcd

dcarley:
Now that the fix for `UpsertServer()` has been merged and we've bumped our
dependency. This makes running the tests easier, because you don't need to
have etcd installed and we don't need to cleanup the datastore before each
test.


fsouza:
Thank you!

PR #1227, opened 2015-07-03
docs/reference/api: Document /healthcheck endpoint

dcarley:
This is good. We should tell people about it.

The trailing slash is intentional per `api/server.go`.


fsouza:
Thank you!

PR #1229, opened 2015-07-06
router/vulcand: Add healthcheck

dcarley:
Which ensures that it gets a successful response from the vulcand API.


fsouza:
Thank you!

Issue #1239, opened 2015-07-16
Potential timeouts while doing platform-add/platform-update


keymon:
Issue description
-----------------------

With the current implementation of the commands to add platforms, `tsuru-admin platform-add` and `tsuru-admin platform-remove`, the process can easily fail due timeouts in Load Balancers or intermediate proxies.

The reason is that tsuru API server will just stream the output of the execution of the `docker build`,  without using websockets or implementing any kind of heartbeat. 

The `docker build` is a process that very likely will hang for a while in some operation without sending any output (e.g. package db update), triggering random timeouts.

You can test this easily with a dockerfile like this:

```
from    ubuntu:14.04
run sleep 120
```

Current solution in other endpoints
--------------------------------------------

Currently  `tsuru-api` implements some short of heartbeat for some endpoints, specifically the `/apps/{appname}/repository/clone`. Basically writes a "Please wait..." every 30 seconds. 

In more detail:
 * API Controller is in this file: https://github.com/tsuru/tsuru/blob/0.11.2/api/server.go
    * For the case of [`/apps/{appname}/repository/clone`](https://github.com/tsuru/tsuru/blob/0.11.2/api/server.go#L166), it calls  [this deploy() function](https://github.com/tsuru/tsuru/blob/0.11.2/api/deploy.go#L23) which [adds wrapper on the HTTP connection](https://github.com/tsuru/tsuru/blob/0.11.2/api/deploy.go#L75) using [io/keepalive_writer.go:NewKeepAliveWriter](https://github.com/tsuru/tsuru/blob/0.11.2/io/keepalive_writer.go#L26)
 * there are two endpoints for the platform, [add](https://github.com/tsuru/tsuru/blob/0.11.2/api/server.go#L158) and [update](https://github.com/tsuru/tsuru/blob/0.11.2/api/server.go#L159), which call [api/platform.go:platformAdd](https://github.com/tsuru/tsuru/blob/0.11.2/api/platform.go#L15) and [api/platform.go:platformUpdate](https://github.com/tsuru/tsuru/blob/0.11.2/api/platform.go#L30)

Potential solutions
-----------------------

We propose several solutions:

 1.  We could add the same logic in the platform add/update related endpoints to avoid timeouts. It it just add a pair or lines of code (we might send a PR soon).

[...]

PR #1240, opened 2015-07-16
Platform Add / Update Heartbeat

RichardKnop:
Installing a platform can take a long time and we
had issues with platform-add/update timing out
after 60 seconds (default Nginx proxy timeout).

We have implemented a simple workaround similar how
deployment endpoint works - using a keep alive writer
and a 30s "please wait" heartbeat.

See also this issue for more thoughts: https://github.com/tsuru/tsuru/issues/1239

### DIFF
--- a/api/platform.go
+++ b/api/platform.go
@@ -7,9 +7,11 @@ package api
 import (
 	"fmt"
 	"net/http"
+	"time"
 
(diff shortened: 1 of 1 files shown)


cezarsa:
Thanks!

Issue #1242, opened 2015-07-16
Inconsistencies in the tsuru API when an object is not found


dhilton:
The tsuru API is a little bit inconsistent when returning the state of an object, if it doesn't exist.

For example, when you try to remove an application that doesn't exist, the API returns a 404.

When you try to remove a user that doesn't exist, the API returns a 500.

My personal preference is that when you try to delete something that doesn't exist it should return a 404, but at least be consistent if possible. Happy to help with a PR if that makes sense.

fsouza:
Hi @dhilton, we should handle 404 in user-remove as well. Please feel free to send the pull request.

Thank you!

fsouza:
@dhilton thanks for reporting and fixing the user-remove inconsistency.

Have you seen this kind of inconsistency in other endpoints as well?

dhilton:
I'm having a look now - we're running a suite of performance tests which create and delete lots of teams, apps and such like - will see if we have any other such issues.

fsouza:
I think we're done with all not founds. I'm closing this issue, @dhilton please feel free to reopen it if you find status code inconsistencies in the API.

Thanks!

PR #1244, opened 2015-07-16
Changed the return code on delete a user to 404 if not found

dhilton:
When querying the user database, if a user is not found the api currently
returns a server error (5XX) which is inconsistent with the other api
calls. This commit updates this behaviour.

I couldn't see any test implications but am happy to update if this is the case.

[#1242]


fsouza:
Thank you!

PR #1282, opened 2015-09-17
Remove deprectated redis driver options

gciotta:
See https://github.com/hipache/hipache/blob/master/lib/cache.js#L14

### DIFF
--- a/docs/installing/hipache-router.rst
+++ b/docs/installing/hipache-router.rst
@@ -43,9 +43,8 @@ In order to install Hipache, just use apt-get:
 Configuring
 ===========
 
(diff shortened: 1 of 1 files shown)


gciotta:
Hi,

I believe the build failed for reasons unrelated to the change, maybe someone can re-trigger it? Also, just to explain the reasoning behind this PR, the options mentioned in the tutorial are not present in the stock config file installed by hipache  anymore, only the `driver` option is there.

fsouza:
Hi @gciotta, thanks for contributing and sorry about the late response. The failure is indeed unrelated, I will merge the pull request.

PR #1289, opened 2015-10-01
Fix URL for documentation in Makefile

ggarnier:
When you miss a tsuru dependency, it shows an invalid URL for the documentation.


andrewsmedina:
@ggarnier thank you!

PR #1291, opened 2015-10-02
Change docs to add set memory in java apps

mdjunior:
Adding docs to environment variable JAVA_MAX_MEMORY (it's equal -Xmx).

Refs: https://docs.oracle.com/cd/E13150_01/jrockit_jvm/jrockit/jrdocs/refman/optionX.html#wp999528


fsouza:
Thank you!

Issue #1297, opened 2015-10-14
cmd: target-add does not create ~/.tsuru


fsouza:
```
% tsuru target-add default "localhost:8080"
Error: open /home/vagrant/.tsuru/targets: no such file or directory
% stat ~/.tsuru
stat: cannot stat '/home/vagrant/.tsuru': No such file or directory
% mkdir ~/.tsuru
% tsuru target-add default "localhost:8080"
New target default -> localhost:8080 added to target list
```

The command should automatically create the directory.

fsouza:
It has been fixed in the PR #1300.

PR #1300, opened 2015-10-15
Creating .tsuru directory before write a target

raelmax:
This patch solve the issue #1297


fsouza:
I've discussed with @raelmax offline, and he will work on two improvements in the PR:

- use os.MkdirAll instead of os.Stat + os.Mkdir
- add a test reproducing the bug and ensure that it passes with the fix

@raelmax thanks for contributing! :D

fsouza:
We're getting ready for a new tsuru-admin release, so I will merge this pull request and do the pending changes.

@raelmax thanks for contirbuting.

PR #1337, opened 2016-01-15
fix license year in all go file

avelino:


tarsisazevedo:
Hi @avelino, thanks for this PR, but we have a policy to just update license in files we change.
We have a script to check that in `misc/check-license.sh`.

PR #1350, opened 2016-01-27
Sleep mode

### LINKED ISSUE #1322: units in sleep mode
The goal is enabled external services put units that are not used to "sleep" and creates a way to these services wake up an unit that is sleeping when necessary.

# Put units to sleep

Sleep should stop unit and change the route to a proxy responsible to wake up the unit.

```bash
$ curl http://mytsuru.com/apps/myapp/sleep?process=web&proxy=http://proxy.com
```

# Wake up units

Should remove the route to the proxy and start unit.

```bash
$ curl http://mytsuru.com/apps/myapp/wakeup?process=web
```

# Points to consider

* Add `sleeping` as unit status
* `routes-rebuild` should not rebuild `sleeping` units
* `app-restart` or `app-start` should wake up a `sleeping` unit

andrewsmedina:
add support for sleep mode:

* added a route that put units to sleep
* added a new unit status "Asleep"

closes #1322


dnsaoki2:
@andrewsmedina thank you

PR #1360, opened 2016-02-11
App list filter by status

ggarnier:
Adds support to filter app list by unit status. You can pass more than one status, separated by `,`:

```
curl -H 'Authorization: b token'  "http://localhost:8080/apps?status=stopped,asleep"
```


andrewsmedina:
@ggarnier thank you!

PR #1362, opened 2016-02-12
Complete docs for add unmanaged node with pool at the setttings

pedrosnk:
Improve the docs when adding unmanaged node


cezarsa:
Thanks!

PR #1381, opened 2016-03-04
app: Fixes duplicated service instance at bind

wpjunior:
The env TSURU_SERVICES in many cases it was duplicated service instances in json.


andrewsmedina:
@wpjunior thank you!

PR #1382, opened 2016-03-09
Support granular permission schemes for deploy

fsouza:
So permission can be controlled on deployment kind as well.

Some use cases:

* supporting only git-based deployments (so users can always track the current commit in production)
* supporting only image-based deployment in certain apps (and use image-based deploy for promoting versions)
* supporting only Dockerfile-based deployment (I don't know why)
* disabling rollback <s>(move fast and break things, that was already supported, sorry)</s>


andrewsmedina:
@fsouza the "drag and drop" deploy will be handled as a "upload" deploy?

fsouza:
@andrewsmedina how does the API identify a drag and drop deploy? Using the origin?

andrewsmedina:
@fsouza yes, the api uses the origin to identify the drag and drop deploy

fsouza:
As we talked offline, just pushed some changes:

1. Abandon #1383 
1. Remove permission for drag-and-drop
1. git-based deployments are only accepted and recognized if the internal token is being used (i.e. Gandalf made the API call)
1. Provide a permission and deployment type for archive-url based deployment (currently a hidden feature of the tsuru API)

andrewsmedina:
@fsouza thank you!

Issue #1394, opened 2016-03-30
Easier way to create a plan


rodrigodealer:
I was setting up a tsuru environment for a presentation in the company that I work in and at some point, I thought that maybe we could be able to customize the unit on the plan creation, something like: 

`$ tsuru-admin plan-create basics -c 2 -m 1G -s 1G -r hipache -d`

And we can use other ways, like M as well.

What do you guys think?

PR #1395, opened 2016-04-01
Supports for '512M' as value for memory and swap

rodrigodealer:
This is related to #1394.

With this change, It should be possible to user to use the following notation in order to create a plan:

`$ tsuru-admin plan-create basics -c 2 -m 1G -s 1G -r hipache -d`

The following options are available:
- K for kilobytes
- M for megabytes
- G for gigabytes

This is my first PR for a Golang project, so I'm not so sure about where the function `getSize ` should be placed, but some feedback is welcome :)

### DIFF
--- a/api/plan.go
+++ b/api/plan.go
@@ -20,7 +20,6 @@ func addPlan(w http.ResponseWriter, r *http.Request, t auth.Token) error {
 	cpuShare, _ := strconv.Atoi(r.FormValue("cpushare"))
 	isDefault, _ := strconv.ParseBool(r.FormValue("default"))
 	memory := getSize(r.FormValue("memory"))
-
 	swap := getSize(r.FormValue("swap"))
 	plan := app.Plan{
 		Name:     r.FormValue("name"),
--- a/api/plan.go
+++ b/api/plan.go
@@ -112,7 +112,6 @@ func getSize(formValue string) int64 {
 		default:
 			return 0
 		}
-	} else {
-		return value
 	}
+	return value
 }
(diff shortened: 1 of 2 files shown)


api/plan.go
```diff
@@ -19,8 +19,9 @@ import (
 func addPlan(w http.ResponseWriter, r *http.Request, t auth.Token) error {
 	cpuShare, _ := strconv.Atoi(r.FormValue("cpushare"))
 	isDefault, _ := strconv.ParseBool(r.FormValue("default"))
-	memory, _ := strconv.ParseInt(r.FormValue("memory"), 10, 64)
-	swap, _ := strconv.ParseInt(r.FormValue("swap"), 10, 64)
+	memory := getSize(r.FormValue("memory"))
+
```
andrewsmedina: can you remove this empty line?

rodrigodealer: Done.

api/plan.go
```diff
@@ -95,3 +96,24 @@ func listRouters(w http.ResponseWriter, r *http.Request, t auth.Token) error {
...
+		case "K":
+			return size * OneKbInBytes
+		case "M":
+			return size * OneKbInBytes * OneKbInBytes
+		case "G":
+			return size * OneKbInBytes * OneKbInBytes * OneKbInBytes
+		default:
+			return 0
+		}
+	} else {
```
andrewsmedina: this `else` is unecessary you can just return the `value`:

```
if err != nil {
   // your code
}
return value
```

rodrigodealer: Good point!

rodrigodealer:
Let me know if you think I should squash my commits.

andrewsmedina:
@rodrigodealer thank you!

PR #1396, opened 2016-04-03
[docs] Fix root-user-create call

jpic:
This fixes:

```
flag provided but not defined: --config
Usage of tsuru flags:
-h, --help  (= false)
    Display help and exit
-v, --verbosity  (= 0)
    Verbosity level: 1 => print HTTP requests; 2 => print HTTP requests/responses
--version  (= false)
    Print version and exit
```

### DIFF
--- a/docs/installing/api.rst
+++ b/docs/installing/api.rst
@@ -136,7 +136,7 @@ For a description
 
 ::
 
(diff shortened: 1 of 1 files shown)


andrewsmedina:
@jpic thank you!

PR #1418, opened 2016-05-19
Ignoring intermittent errors on EC2 api

guilhermef:
When starting  a node, a few times I get a 503 from the EC2 api,
this would stop the instance start.
Now It retries the api call before a hard timeout.

### DIFF
--- a/iaas/ec2/task.go
+++ b/iaas/ec2/task.go
@@ -55,14 +55,19 @@ func (t *ec2WaitTask) Run(job monsterqueue.Job) {
 	var notifiedSuccess bool
 	t0 := time.Now()
 	for {
+		if time.Now().Sub(t0) > time.Duration(2*timeout)*time.Second {
+			job.Error(errors.New("hard timeout"))
+			break
+		}
 		log.Debugf("ec2: waiting for dnsname for instance %s", machineId)
 		input := ec2.DescribeInstancesInput{
 			InstanceIds: []*string{aws.String(machineId)},
 		}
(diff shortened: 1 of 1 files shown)


iaas/ec2/task.go
```diff
@@ -55,14 +55,19 @@ func (t *ec2WaitTask) Run(job monsterqueue.Job) {
...
+		}
 		log.Debugf("ec2: waiting for dnsname for instance %s", machineId)
 		input := ec2.DescribeInstancesInput{
 			InstanceIds: []*string{aws.String(machineId)},
 		}
 		resp, err := ec2Inst.DescribeInstances(&input)
 		if err != nil {
-			job.Error(err)
-			break
+			log.Debugf("ec2: api error")
```
tarsisazevedo: You can use `log.Debug` here.

iaas/ec2/task.go
```diff
@@ -55,14 +55,19 @@ func (t *ec2WaitTask) Run(job monsterqueue.Job) {
 	var notifiedSuccess bool
 	t0 := time.Now()
 	for {
+		if time.Now().Sub(t0) > time.Duration(2*timeout)*time.Second {
```
tarsisazevedo: if `timeout` is 0 there is a chance that task will not run. Is this a expected behavior?

guilhermef: There's a scenario where timeout is 0 ?

tarsisazevedo: No, sorry 😄

tarsisazevedo:
Thanks! \o/

PR #1439, opened 2016-08-12
Update gandalf.rst

cqteixeira:
Adding the information about mongodb server. Most people have problem at this steps because they receive the message "No reachable servers" on tsuru-api and have no idea that the problem is with the communication between gandalf and mongodb.


andrewsmedina:
@cqteixeira thank you!

PR #1443, opened 2016-08-24
Optmizing client update warn

nelsonmarcos:


andrewsmedina:
I'm -1 for this change. This PR makes the warning message to big and is not specific by client and platform and is useful only for brew and apt users.

cezarsa:
I agree with @andrewsmedina. The installation instructions are already described in http://docs.tsuru.io/en/latest/using/install-client.html which is linked in the current warning. If we merge this we would have to remember to always update the same instructions in two different places and this could lead to inconsistencies. It also doesn't scale as the message would only get bigger as more platforms were added.

I'm closing this for now. Feel free to comment if you have any other suggestion on how to improve this.

Issue #855, opened 2014-07-16
tsuru run --isolated: creating a container just to run a command


magnotorres:
Would be a nice feature if we can use tsuru run "command" --flag (--isolated?) that creates a new unit (like tsuru add-unit) without add it to the router, and that will exist until the command finishes
Doing that, I can run any command without compete with any running instance of the current application

andrestc:
Closed by #1478

PR #1478, opened 2016-10-14
Implements isolated flag to app-run command

guilhermebr:
This PR solves the issue #855 implementing the flag --isolated to app-run command.
If accepted user will can execute command in an ephemeral container.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -866,7 +866,7 @@ func (app *App) InstanceEnv(name string) map[string]bind.EnvVar {
 
 // Run executes the command in app units, sourcing apprc before running the
 // command.
-func (app *App) Run(cmd string, w io.Writer, once bool) error {
+func (app *App) Run(cmd string, w io.Writer, once, isolated bool) error {
 	if !app.available() {
 		return stderr.New("App must be available to run commands")
 	}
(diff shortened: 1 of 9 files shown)


app/app.go
```diff
@@ -866,25 +866,25 @@ func (app *App) InstanceEnv(name string) map[string]bind.EnvVar {
 
 // Run executes the command in app units, sourcing apprc before running the
 // command.
-func (app *App) Run(cmd string, w io.Writer, once bool) error {
+func (app *App) Run(cmd string, w io.Writer, once, isolated bool) error {
```
cezarsa: One small nitpick, it might be better to create a new type, something like RunArgs, to group this parameters and send them along the way to `sourced` and `run`. Having two booleans in the signature may start getting messy in the future and lead to bugs where the two are swapped by mistake.

provision/provisiontest/fake_provisioner.go
```diff
@@ -1068,6 +1068,36 @@ func (p *FakeProvisioner) ExecuteCommandOnce(stdout, stderr io.Writer, app provi
 	return nil
 }
 
+func (p *FakeProvisioner) ExecuteCommandIsolated(stdout, stderr io.Writer, app provision.App, cmd string, args ...string) error {
```
cezarsa: Could you add a simple test for this function in fake_provisioner_test.go? Similar to https://github.com/tsuru/tsuru/blob/master/provision/provisiontest/fake_provisioner_test.go#L871. Just to make sure this code is executed during tests.

guilhermebr:
@cezarsa PTAL

andrewsmedina:
@guilhermebr thank you!

PR #1507, opened 2016-12-05
fix link to docs of docker-machine drivers

luzfcb:


ArxdSilva:
Thank you for contributing @luzfcb!

Issue #1523, opened 2017-01-16
Digitalocean droplet is not effectively removed


felipeweb:
To remove a digitalocean droplet we must shutdown before remove.

I'm fixing this.

PR #1524, opened 2017-01-16
shutdown droplet before remove

felipeweb:
#1523


felipeweb:
i will fix the tests, i`m trying to understand why my changes causes a EOF error

andrestc:
When calling `Shutdown` and `PowerOff` the DO client tries to [decode the response body](https://github.com/tsuru/tsuru/blob/master/vendor/github.com/digitalocean/godo/godo.go#L262). You should return some content on the [`fakeserver`](https://github.com/tsuru/tsuru/blob/master/iaas/digitalocean/iaas_test.go#L117) for these endpoints.

felipeweb:
thank you for help @andrestc  i open other PR #1525

Issue #1527, opened 2017-01-19
Add subcommands to tsuru-client


andrestc:
Commands from `tsuru-admin` and `crane` are being merged into `tsuru-client` on version 1.2. Currently, when running `tsuru`, the user has around 140 commands to choose from. I think this can pollute the user experience.

I think we could try to split our commands into subcommands.

One possible way of grouping would be to add the `manage` subcommand to aggregate all commands that originated from `tsuru-admin` and `crane` and are mostly used by tsuru admins and service admins. This would already bring the command count (when issuing `tsuru help`) to around 80 possible commands. To run a `tsuru node-add` cmd, one would run `tsuru manage node-add`.

Another possible approach would be to group commands by targets (app, service, node, event... etc), e.g, `tsuru app list/create/remove/deploy`, `tsuru event list/info/cancel`. I think this is a more natural grouping and seems to be an approach commonly adopted by other clients, e.g, `docker service create/rm/ls...`.

Regardless of the approach adopted, we should support both ways (current and the next one) with a deprecation warning and remove on the following version.

guilhermebr:
I agree with both changes.

separated commands are more natural and intuitive, and will decrease considerably the number of commands to list.

and group both service and admin commands into `manage` subcommand will prevent mistakes.

cezarsa:
I like the idea of grouping by target, once upon a time I even tinkered with this idea and I can bring this code back. It had the added benefit of maintaining backwards compatibility with the current dash separated commands.

One thing to consider is that our current CLI API already has the concept of topics (try `tsuru help target`). This idea could be expanded with the creation of a topic for each target and adding documentation for each new topic as we go.

andrestc:
We could bring that code back (I would like to maintain the compatibility for at least one version, I don't think we need to keep this forever).

I like the idea of having a help topic for each target (adding their related subcommands etc).

PR #1531, opened 2017-01-25
cmd: implicit topics help and allow space as separator

cezarsa:
Creating a pull-request for discussion related to #1527.

With these changes, the first part of any dash separated command will create an implicit topic on tsuru. Which we can review and add proper descriptions latter.

Building tsuru-client with these changes result in the following behavior:

```
$ tsuru help
tsuru version 1.2.0-dev.

Usage: tsuru command [args]

Available commands:
  app-create                   Creates a new app using the given name and platform
  ... ALL STILL COMMANDS LISTED .. 

Use tsuru help <commandname> to get more information about a command.

Available topics:
  app                  
  bs                   
  certificate          
  cname                
  containers           
  docker               
  env                  
  event                
  install              
  key                  
  machine              
  node                 
  plan                 
  platform             
  plugin               
  pool                 
  role                 
  service              
  target               Target is used to manage the address of the remote tsuru server
  team                 
  token                
  unit                 
  user                 

Use tsuru help <topicname> to get more information about a topic.
```

```
$ tsuru app
The following commands are available in the "app" topic:

  app-create           Creates a new app using the given name and platform
  ... ALL `app-` COMMANDS...

Use tsuru help <commandname> to get more information about a command.
```

```
$ tsuru app list
# WORKS
$ tsuru app-list
# ALSO WORKS
```

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -79,7 +79,7 @@ func BuildBaseManager(name, version, versionHeader string, lookup Lookup) *Manag
 	m.Register(&targetRemove{})
 	m.Register(&targetSet{})
 	m.Register(userInfo{})
-	m.RegisterTopic("target", fmt.Sprintf(targetTopic, name))
+	m.RegisterTopic("target", targetTopic)
 	return m
 }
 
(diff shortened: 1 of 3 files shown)


fsouza:
magic

avelino:
gandalf_to_the_music

andrestc:
I think we just need to figure out how we gonna handle the `tsuru install` case and im +1 for merging this! We should update the documentation regarding the new syntax  and also add documentation for each topic for the 1.3 release

PR #1533, opened 2017-01-25
update Copyright year 2016 to 2017

avelino:
was done using `sed`


andrestc:
Hi @avelino, thanks for this but we decided to keep the file creation date at the copyright headers. This seems to be the most common in other projects under BSD license.

Issue #1539, opened 2017-02-07
Update machine-template IaaS


andrestc:
We should support a flag, e.g, `-i iaas_name` to provide a way for the user to update the iaas of a template when running `tsuru machine-template-update`. This will help people that decide to migrate from our own iaas implementation to docker-machine drivers.

PR #1544, opened 2017-02-24
Iaas template update (WIP)

ArxdSilva:
this was ment to solve tsuru/tsuru#1539

### DIFF
--- a/api/iaas.go
+++ b/api/iaas.go
@@ -6,6 +6,7 @@ package api
 
 import (
 	"encoding/json"
+	"fmt"
 	"net/http"
 
 	"github.com/ajg/form"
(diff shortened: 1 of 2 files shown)


api/iaas.go
```diff
@@ -225,13 +225,17 @@ func templateUpdate(w http.ResponseWriter, r *http.Request, token auth.Token) (e
 		return &errors.HTTP{Code: http.StatusBadRequest, Message: err.Error()}
 	}
 	templateName := r.URL.Query().Get(":template_name")
+	iaasName := r.Form.Get("IaaSName")
```
andrestc: You should check if the `iaasName` iaas a valid iaas name (it exists).

api/iaas.go
```diff
@@ -225,13 +225,17 @@ func templateUpdate(w http.ResponseWriter, r *http.Request, token auth.Token) (e
...
 	dbTpl, err := iaas.FindTemplate(templateName)
 	if err != nil {
 		if err == mgo.ErrNotFound {
 			return &errors.HTTP{Code: http.StatusNotFound, Message: "template not found"}
 		}
 		return err
 	}
+	if (dbTpl.IaaSName != iaasName) && (iaasName != "") {
```
andrestc: I think you don't need the first check: `(dbTpl.IaaSName != iaasName)`.

andrestc:
Left some commends, I think we should add some tests for this.

api/iaas.go
ArxdSilva: It already does [here](https://github.com/tsuru/tsuru/blob/master/iaas/iaas.go#L115-L124) after returning `dbTpl.Update`

api/iaas.go
```diff
@@ -148,23 +150,27 @@ func templateCreate(w http.ResponseWriter, r *http.Request, token auth.Token) (e
...
 	}
-	evt, err := event.New(&event.Opts{
-		Target:     event.Target{Type: event.TargetTypeIaas, Value: paramTemplate.IaaSName},
-		Kind:       permission.PermMachineTemplateCreate,
-		Owner:      token,
-		CustomData: event.FormToCustomData(r.Form),
-		Allowed:    event.Allowed(permission.PermMachineReadEvents, iaasCtx),
-	})
+	_, err = iaas.FindTemplate(paramTemplate.Name)
 	if err != nil {
```
andrestc: I would try to be a little more carefull here, something like:
```go
if err != nil {
    if err == mgo.ErrNotFound {
        return &errors.HTTP{Code: http.StatusConflict, Message: fmt.Sprintf("template name \"%s\" already used", paramTemplate.Name)}
    }
    return err
}
... the rest of the code here.
```

api/iaas.go
```diff
@@ -148,23 +150,27 @@ func templateCreate(w http.ResponseWriter, r *http.Request, token auth.Token) (e
 	if !allowed {
 		return permission.ErrUnauthorized
 	}
-	evt, err := event.New(&event.Opts{
```
andrestc: I think we could create the event before checking for the template existence. I think it's good to have this failure recorded on the events.

andrestc:
LGTM

Issue #1502, opened 2016-11-25
Blockable events


andrestc:
It should be possible to prevent some events from running. There are many scenarios where blocking some events could be useful, e.g, blocking deploys during a `tsurud` maintenance or upgrade.

The UX would be something like: `tsuru event-block -k app.deploy`. We could support other kind of filters for more granular blocking (blocking `unit-add` on some specific pools due to resources limitation for example) and maybe support setting a ttl for the duration of the block.

Attempts to run a blocked event would be still created and added to the `event-list` with a `blocked` status and, when adding a block, the admin should add a `reason` message to be displayed to the user that attempted to run an action that triggered a blocked event.

PR #1551, opened 2017-03-10
Blockable events

andrestc:
This PR adds support for blocking events (#1502). Blocked events will appear on the event list as erroed.


event/block.go
```diff
@@ -0,0 +1,131 @@
...
+
+func checkIsBlocked(evt *Event) error {
+	query := bson.M{"$and": []bson.M{
+		{"active": true},
+		{"$or": []bson.M{{"kindname": evt.Kind.Name}, {"kindname": ""}}},
+		{"$or": []bson.M{{"ownername": evt.Owner.Name}, {"ownername": ""}}},
+		{"$or": []bson.M{
+			{"target": evt.Target},
+			{"target": bson.M{"$exists": false}},
+			{"$and": []bson.M{{"target.type": evt.Target.Type}, {"target.value": ""}}}}},
```
cezarsa: I think this last `$and` isn't needed, it could be `{"target.type": evt.Target.Type, "target.value": ""}`.

api/event.go
```diff
@@ -9,6 +9,8 @@ import (
 	"fmt"
 	"net/http"
```
andrewsmedina: this empty line should be removed

Issue #1550, opened 2017-03-08
Enable tagging applications/service-instances


andrestc:
One should be able to add "tags" to apps/service-instances which are simply free text metadata about that item. These should be shown on `app-info/service-instance-info` and be available for filtering when listing apps/services-instances.

PR #1553, opened 2017-03-13
App tags

ggarnier:
Adds support to tags in apps (#1550). Tags can be set in app creation, or edited in update. When the tags param is set, it replaces all current tags. When it's omitted, the original tags are kept.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -372,7 +372,7 @@ func (app *App) Update(updateData App, w io.Writer) (err error) {
 			}
 		}()
 	}
-	if len(tags) > 0 {
+	if tags != nil {
 		app.Tags = tags
 	}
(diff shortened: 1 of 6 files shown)


app/app.go
```diff
@@ -394,6 +400,22 @@ func (app *App) Update(updateData App, w io.Writer) (err error) {
...
 }
 
+func processTags(tags []string) []string {
+	if tags == nil {
+		return nil
+	}
+	processedTags := []string{}
+	usedTags := make(map[string]bool)
+	for _, tag := range tags {
+		tag = strings.Trim(tag, " ")
```
cezarsa: You can use `strings.TrimSpace` here. It also handles \n and other space chars.

PR #1561, opened 2017-03-28
Rollback update, naming improvement & removal of unused code

### LINKED ISSUE #1414: [Feature] Remove old deploy
When deploying an app, old deploys are keeped for application rollback.

But sometimes these deploys even if they've passed the build process do not work. Can we add a `tsuru app-deploy-remove` command to remove old deploys. This can help to not rollback on a bugged version and can help optimize hdd space on servers.

ArxdSilva:
This PR adds the endpoint to rollback update, allows the user to update a image's so It can be rollbackable or not, add a message into the image and also provides a better naming to some functions in image's package.

fixes #1414


app/deploy.go
```diff
@@ -244,6 +239,14 @@ func Deploy(opts DeployOptions) (string, error) {
 	return imageId, nil
 }
 
+func RollbackUpdate(appName, imgId, reason string, rollback bool) error {
+	err := image.ValidateAppImage(appName, imgId)
```
cezarsa: This will probably always fail. This happens because the imgId here most likely isn't a full image id, it's just the version suffix. The logic for transforming the received suffix into a full image id currently is [here in the Deploy method](https://github.com/tsuru/tsuru/pull/1561/files#diff-975107ae4993861af2c7808681346ecdR208). You should probably extract it to a function and use it in both places.

app/image/image.go
```diff
@@ -303,6 +305,25 @@ func DeleteAllAppImageNames(appName string) error {
 	return coll.RemoveId(appName)
 }
 
+func UpdateAppImageRollback(appName, img, reason string, rollback bool) error {
+	dataColl, err := imageCustomDataColl()
+	if err != nil {
+		return err
+	}
+	defer dataColl.Close()
+	err = dataColl.Update(bson.M{"_id": fmt.Sprintf("%s:%s", appBasicImageName(appName), img)},
```
cezarsa: The received image id here should already be a full image id (see previous comment on app.RollbackUpdate). So there's no need for this call to `fmt.Sprintf` simply filter with `{"_id": img}`.

app/image/suite_test.go
```diff
@@ -0,0 +1,35 @@
+package image
```
cezarsa: There is already a test suite declared on this package. It's inside the `app/image/image_test.go` file. You don't need this file.

app/deploy_test.go
```diff
@@ -869,3 +869,25 @@ func (s *S) TestRebuild(c *check.C) {
...
+	c.Check(dbMetadata.DisableRollback, check.Equals, true)
+	c.Check(dbMetadata.Reason, check.Equals, "buggy version")
+	err = RollbackUpdate("myapp", "v1", "", false)
+	c.Check(err, check.IsNil)
+	dbMetadata, err = image.GetImageMetaData(data.Name)
+	c.Check(err, check.IsNil)
+	c.Check(dbMetadata.DisableRollback, check.Equals, false)
+	c.Check(dbMetadata.Reason, check.Equals, "")
+	err = RollbackUpdate("myapp", "v10", "", false)
+	c.Check(err, check.NotNil)
```
cezarsa: Use `check.ErrorMatches` to basically validate the error message being returned.

app/deploy_test.go
```diff
@@ -869,3 +869,25 @@ func (s *S) TestRebuild(c *check.C) {
 	c.Assert(writer.String(), check.Equals, "Rebuild deploy called")
 	c.Assert(imgID, check.Equals, "app-image")
 }
+
+func (s *S) TestRollbackUpdate(c *check.C) {
```
cezarsa: You should test more scenarios (or add more test funcs) like enabling the image after it's been disabled.

api/deploy.go
```diff
@@ -125,14 +125,12 @@ func deploy(w http.ResponseWriter, r *http.Request, t auth.Token) (err error) {
 		Build:      build,
 		Message:    message,
 	}
-	opts.GetKind()
```
ArxdSilva: Yes, that was mistaken

app/image/suite_test.go
ArxdSilva: I moved to this file the `suite_test` part from `app/image/image_test.go`.

cezarsa: Oh okay then for some reason github collapsed the changes to image_test.go and I missed them.

app/image/image_test.go
```diff
@@ -2,231 +2,203 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-package image_test
```
cezarsa: If you remove the `_test` suffix here, you must also remove the import for `"github.com/tsuru/tsuru/app/image"`. Otherwise this won't even compile.

ArxdSilva: It's already done, the import was a mistake made while merging the master fork

api/deploy.go
```diff
@@ -400,3 +399,77 @@ func deployRebuild(w http.ResponseWriter, r *http.Request, t auth.Token) error {
...
+		App:  instance,
+		User: t.GetUserName(),
+		Kind: app.DeployRollback,
+	}
+	var imageID string
+	evt, err := event.New(&event.Opts{
+		Target:        appTarget(appName),
+		Kind:          permission.PermAppUpdate,
+		Owner:         t,
+		CustomData:    opts,
```
andrestc: I think this `CustomData` should not be an `app.DeployOptions`, It should probably be created from the request form values with `event.FormToCustomData`

api/deploy.go
```diff
@@ -400,3 +399,77 @@ func deployRebuild(w http.ResponseWriter, r *http.Request, t auth.Token) error {
...
+		Allowed:       event.Allowed(permission.PermAppReadEvents, contextsForApp(instance)...),
+		AllowedCancel: event.Allowed(permission.PermAppUpdateEvents, contextsForApp(instance)...),
+		Cancelable:    true,
```
andrestc: I don't think this event needs to be cancelable

api/deploy.go
```diff
@@ -400,3 +399,77 @@ func deployRebuild(w http.ResponseWriter, r *http.Request, t auth.Token) error {
...
+	})
+	if err != nil {
+		return err
+	}
+	defer func() { evt.DoneCustomData(err, map[string]string{"image": imageID}) }()
```
andrestc: not needed once we add this to the CustomData when creating the event

api/server.go
```diff
@@ -153,6 +153,7 @@ func RunServer(dry bool) http.Handler {
 	logPostHandler := AuthorizationRequiredHandler(addLog)
 	m.Add("1.0", "Post", "/apps/{app}/log", logPostHandler)
 	m.Add("1.0", "Post", "/apps/{appname}/deploy/rollback", AuthorizationRequiredHandler(deployRollback))
+	m.Add("1.0", "Put", "/apps/{appname}/deploy/rollback/update", AuthorizationRequiredHandler(deployRollbackUpdate))
```
andrestc: this should be added to `1.3` instead of `1.0`

api/deploy.go
```diff
@@ -400,3 +399,77 @@ func deployRebuild(w http.ResponseWriter, r *http.Request, t auth.Token) error {
...
+		return &tsuruErrors.HTTP{
+			Code:    http.StatusBadRequest,
+			Message: "Reason cannot be empty while enabling rollback",
+		}
+	}
+
+	err = app.RollbackUpdate(appName, img, reason, rollback)
+	if err != nil {
+		return &tsuruErrors.HTTP{
+			Code:    http.StatusNotFound,
```
andrestc: I dont think all errors from `app.RollbackUpdate` should be treated as `http.StatusNotFound`

ArxdSilva: I think `http.StatusConflict` would be better, what do you think?

api/deploy.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2013 tsuru authors. All rights reserved.
```
andrestc: This should be kept as 2013 (the year when the file was created).

app/deploy.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2014 tsuru authors. All rights reserved.
```
andrestc: Same as commented before.

app/deploy_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2014 tsuru authors. All rights reserved.
```
andrestc: Same as commented before.

app/image/image.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2014 tsuru authors. All rights reserved.
```
andrestc: Same as commented before.

provision/provisiontest/fake_provisioner.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2012 tsuru authors. All rights reserved.
```
andrestc: Same as commented before.

api/deploy.go
```diff
@@ -400,3 +400,71 @@ func deployRebuild(w http.ResponseWriter, r *http.Request, t auth.Token) error {
...
+		return &tsuruErrors.HTTP{
+			Code:    http.StatusBadRequest,
+			Message: fmt.Sprintf("Status `enabled` set to: %s instead of `true` or `false`", rb),
+		}
+	}
+	reason := r.FormValue("reason")
+	if (reason == "") && (rollback == false) {
+		return &tsuruErrors.HTTP{
+			Code:    http.StatusBadRequest,
+			Message: "Reason cannot be empty while enabling rollback",
```
andrestc: I think this should be "while disabling rollback" instead of "enabling".

api/deploy.go
```diff
@@ -400,3 +400,71 @@ func deployRebuild(w http.ResponseWriter, r *http.Request, t auth.Token) error {
...
+		Cancelable:    false,
+	})
+	if err != nil {
+		return err
+	}
+	defer func() { evt.Done(err) }()
+	err = app.RollbackUpdate(instance, img, reason, rollback)
+	if err != nil {
+		return &tsuruErrors.HTTP{
+			Code:    http.StatusConflict,
```
andrestc: I dont think http.StatusConflict is the right choice here, e.g, if the image is not found we should probably return a BadRequest and for generic errors we should return InternalServerError.

api/deploy.go
```diff
@@ -461,9 +464,17 @@ func deployRollbackUpdate(w http.ResponseWriter, r *http.Request, t auth.Token)
 	defer func() { evt.Done(err) }()
 	err = app.RollbackUpdate(instance, img, reason, rollback)
 	if err != nil {
-		return &tsuruErrors.HTTP{
-			Code:    http.StatusConflict,
-			Message: err.Error(),
+		switch err {
+		case image.ErrInvalidVersion:
+			return &tsuruErrors.HTTP{
+				Code:    http.StatusInternalServerError,
```
andrestc: I think the errors here are backwards: first case should be `http.StatusBadRequest`. You actually do not need the second case (you can just return the error directly and it will be represented as a ` http.StatusInternalServerError` by our [error handling middleware](https://github.com/tsuru/tsuru/blob/456a675a3aba0163195d66c14688f9a59c2c7b6c/api/middleware.go#L102).

app/image/image.go
```diff
@@ -19,6 +19,13 @@ import (
 	"gopkg.in/mgo.v2/bson"
 )
 
+var (
+	ErrImageNotFound     = errors.New("Image not found")
```
andrestc: Maybe we could define these as structs so we can keep the context? (what image and what version were not found?)

Check this example [here](https://github.com/tsuru/tsuru/blob/master/provision/provision.go#L33).

app/deploy_test.go
```diff
@@ -869,3 +863,28 @@ func (s *S) TestRebuild(c *check.C) {
 	c.Assert(writer.String(), check.Equals, "Rebuild deploy called")
 	c.Assert(imgID, check.Equals, "app-image")
 }
+
+func (s *S) TestRollbackUpdate(c *check.C) {
```
andrestc: I think you could add more tests to this func (the same cases your did for the handler, but calling app.RollBackUpdate directly) as @cezarsa. The rest lgtm

andrestc:
Added a single commit (reinforcing the need to add more tests that call app.RollBackUpdate directly)

provision/docker/provisioner.go
```diff
@@ -368,19 +368,16 @@ func (p *dockerProvisioner) Sleep(app provision.App, process string) error {
...
-			valid = true
-			break
-		}
+	imgMetaData, err := image.GetImageMetaData(imageID)
+	if err != nil {
+		return "", err
 	}
-	if !valid {
-		return "", errors.Errorf("Image %q not found in app", imageID)
+	if imgMetaData.DisableRollback {
```
cezarsa: What do you think about moving this check to `func Deploy` in `app/deploy.go`? This way we wouldn't have to worry about adding this same logic to every provisioner.
Also I don't think there's need to call `GetAppImageBySuffix` here, the app package will have done this already and the imageID here would be a full valid imageID not a suffix.

PR #1566, opened 2017-04-13
Builder Package

guilhermebr:
This PR creates Builder package

Builder package will be responsible for image building stage of deploy, 
removing all builder logics from provisioners.

This PR also implements docker builder, responsible for building docker images 
for docker-based provisioners.

With this PR provision will only have one Deploy method,
upload, archive, image and rebuild deploy will be created at builder stage.


builder/builder_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+func (s S) TestRegistry(c *check.C) {
+	var b1, b2 Builder
+	Register("my-builder", b1)
+	Register("your-builder", b2)
+	builders, err := Registry()
+	c.Assert(err, check.IsNil)
+	alt1 := []Builder{b1, b2}
+	alt2 := []Builder{b2, b1}
+	if !reflect.DeepEqual(builders, alt1) && !reflect.DeepEqual(builders, alt2) {
```
ggarnier: Isn't this the same as `c.Assert(builders, check.Not(check.DeepEquals(alt1)))` and `c.Assert(builders, check.Not(check.DeepEquals(alt2)))`?

Issue #1430, opened 2016-07-12
User-list should display information of the user that is logged in


raulgbrmf:
If we create an non-admin user and then log-in with it, when issuing the user-list command it returns only the header : 
+------+-------+
 |  User | Roles  |
+------+-------+

It should return at least informations regarding the user that is logged in.

mateuspontes:
There are many ways to resolve this, I think the best solution is to check Roles in ListUsers() before the loop, but this logic is owned by createAPIUser. I created a test case and push some code inside the createAPIUser to solve that too, maybe not the best way.

ArxdSilva:
@mateusdelbianco thanks for helping, but when [tsuru-client](https://github.com/tsuru/tsuru-client/blob/0928a3ab9c64d116e0e0fbe49fad9330fe3572c3/tsuru/client/auth.go#L517) makes the GET request on the API, It responds [here](https://github.com/tsuru/tsuru/blob/master/api/server.go#L201) and gets the data from DB [here](https://github.com/tsuru/tsuru/blob/master/api/auth.go#L712). So what I think is best for this problem is checking this `users` and If empty we return the own user's info request.

PR #1576, opened 2017-05-10
User-list bug fix

ArxdSilva:
Fixes bug on Issue #1430 where in a situation that a user has only basic permissions and is not assigned to any teams, using `tsuru user-list` command would not show anything other than the headers.


andrestc:
LGTM, thanks

Issue #1575, opened 2017-05-09
Removing pools with apps


ggarnier:
`pool-remove` command allows removal of pools that have apps. I think we should check this first and show an error message. And maybe add a flag to remove the pool and its apps.

This behavior also creates some side effects:

- `app-list` and `app-info` start to show an error message: `Error: Pool does not exist.`
- you can't remove apps whose pool has been removed

PR #1580, opened 2017-05-23
api: Blocks removal of pools with apps

ArxdSilva:
This blocks the user from removing a pool that still
    contains apps. Fixes tsuru/tsuru#1575

### DIFF
--- a/api/pool.go
+++ b/api/pool.go
@@ -166,7 +166,7 @@ func removePoolHandler(w http.ResponseWriter, r *http.Request, t auth.Token) (er
 		return err
 	}
 	if len(apps) > 0 {
-		return &terrors.HTTP{Code: http.StatusForbidden, Message: "This pool still have apps, before deleting It's needed to migrate or delete all apps"}
+		return &terrors.HTTP{Code: http.StatusForbidden, Message: "This pool has apps, you need to migrate or remove them before removing the pool"}
 	}
 	evt, err := event.New(&event.Opts{
 		Target:     event.Target{Type: event.TargetTypePool, Value: poolName},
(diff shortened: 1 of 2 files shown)


api/pool.go
```diff
@@ -158,6 +159,16 @@ func removePoolHandler(w http.ResponseWriter, r *http.Request, t auth.Token) (er
 		return permission.ErrUnauthorized
 	}
 	poolName := r.URL.Query().Get(":name")
+	filter := &app.Filter{}
+	filter.Pool = poolName
+	contexts := permission.ContextsForPermission(t, permission.PermAppRead)
+	apps, err := app.List(appFilterByContext(contexts, filter))
```
andrestc: I think we could use a empty slice of contexts here (instead of getting the contexts from the user token) so we also take into account apps that the user has no permission on.

api/pool.go
```diff
@@ -158,6 +159,15 @@ func removePoolHandler(w http.ResponseWriter, r *http.Request, t auth.Token) (er
...
 	}
 	poolName := r.URL.Query().Get(":name")
+	filter := &app.Filter{}
+	filter.Pool = poolName
+	apps, err := app.List(appFilterByContext([]permission.PermissionContext{}, filter))
+	if err != nil {
+		return err
+	}
+	if len(apps) > 0 {
+		return &terrors.HTTP{Code: http.StatusForbidden, Message: "This pool still have apps, before deleting It's needed to migrate or delete all apps"}
```
andrestc: I think "This pool has apps, you need to migrate or remove them before removing the pool" reads better. What do you think?

andrestc:
looks good, except for the error message (left a comment there)

andrestc:
LGTM, thanks!

PR #1584, opened 2017-05-25
api: print remove service instance messages in different lines

### LINKED ISSUE #1583: client: API + client messages attached on `service-instance-remove`
When removing a service instance, the messages from the API and client get attached and should be printed in different lines.
screen shot 2017-05-25 at 2 25 19 pm

MayconDionisio:
it fixes #1583


andrestc:
LGTM, thanks

Issue #1379, opened 2016-03-03
Tsuru should have a role-update command.


raulgbrmf:
Now that it is possible to add a description to newly created roles, it would be necessary to create a way to update older roles with a description, or change the description of a role.
Without this possibility, the only way we can do it is by deleting and recreating the role, and this is not well recommended, since all the users assigned to this role will lose their permission associated to it.

PR #1585, opened 2017-05-31
role-update command

ArxdSilva:
Related to #1379

### DIFF
--- a/api/permission.go
+++ b/api/permission.go
@@ -639,7 +639,7 @@ func listDefaultRoles(w http.ResponseWriter, r *http.Request, t auth.Token) erro
 	return json.NewEncoder(w).Encode(roles)
 }
 
-// title: list default roles
+// title: updates a role
 // path: /roles
 // method: PUT
 // responses:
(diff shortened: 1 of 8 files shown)


api/permission.go
```diff
@@ -638,3 +638,40 @@ func listDefaultRoles(w http.ResponseWriter, r *http.Request, t auth.Token) erro
 	w.Header().Set("Content-Type", "application/json")
 	return json.NewEncoder(w).Encode(roles)
 }
+
+// title: list default roles
```
andrestc: should be something like "update role"

auth/user.go
```diff
@@ -277,6 +277,48 @@ func (u *User) AddRole(roleName string, contextValue string) error {
...
+	if ctx != "" {
+		role.ContextType, err = permission.ParseContext(ctx)
+		if err != nil {
+			return err
+		}
+	}
+	if desc != "" {
+		role.Description = desc
+	}
+	if newName == "" {
```
andrestc: should this be `newName == "" || newName == name`?

ggarnier:
Maybe. If something goes wrong after step 2, users would have no roles (old or new). In your proposal, users would have both old and new roles.

andrestc:
I think that is probably safer

ArxdSilva:
Did some improvements to what @andrestc suggested, I was ignoring the error on `user.AddRole()`, not anymore. Also moved `role.Name = newRoleName` to where It's really needed.

Issue #1572, opened 2017-04-26
HTTP API based Router


andrestc:
This issue is to track the definition of a base router API contract and the addition of a generic api based router implementation on tsuru api. These will make easier to add new routers (their specific code will live outside tsuru code such as a [TCP router](https://github.com/tsuru/tsuru/issues/983).

The conf for setting up a router would be such as:

``` yml
routers:
  my-generic-router:
    type: generic # (or api?...)
    endpoint: http://my-router-endpoint/v1
```

The generic router type expects the endpoint to implement a certain api contract to be able to provide all the Router interface functionality (we will need to rethink show we check for specific functionality, such [as TLS](https://github.com/tsuru/tsuru/blob/master/app/app.go#L1808))

We have also discussed and agreed on removing our `fusis` implementation and moving it a separated repository that will implement the router http api.

PR #1586, opened 2017-06-05
Adds generic router api support and drop fusis

andrestc:
This PR adds support for generic routers that are HTTP api based. We also decided to drop fusis support. If we decide to use fusis in the future, we should build an API that matches our contract and plug it as an api router.

Closes: #1572


ggarnier:
Maybe it would be better to replace `setRoutes` method with `addRoute` and `removeRoute`. What do you think, @cezarsa?

router/api/router_test.go
```diff
@@ -433,7 +434,33 @@ func (f *fakeRouterAPI) setRoutes(w http.ResponseWriter, r *http.Request) {
...
+		backend.addresses = append(backend.addresses, req.Addresses...)
+		return
+	}
+	w.WriteHeader(http.StatusNotFound)
+}
+
+func (f *fakeRouterAPI) removeRoutes(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	name := vars["name"]
+	if backend, ok := f.backends[name]; ok {
```
ggarnier: I think you could invert this logic to improve the code readability:

```
backend, ok := f.backends[name]
if !ok {
    w.WriteHeader(http.StatusNotFound)
    return
}
req := &routesReq{}
...
```

PR #1588, opened 2017-06-06
api/node: disables the need for machine perms

andrestc:
This PR disables the need for `PermMachineCreate` and `PermMachineDelete` from adding and removing nodes. `PermMachineCreate` was removed since it wasn't used anywhere.


api/node.go
```diff
@@ -191,14 +184,6 @@ func removeNodeHandler(w http.ResponseWriter, r *http.Request, t auth.Token) (er
 		return permission.ErrUnauthorized
 	}
 	removeIaaS, _ := strconv.ParseBool(r.URL.Query().Get("remove-iaas"))
```
guilhermebr: Just a minor thing about readability... you're using `removeIaaS`  ~20 lines later

PR #1590, opened 2017-06-08
adds support for image-reset on next deploy by using app-update

### LINKED ISSUE #813: app: tsuru command to force rebuild app image
App actually is rebuild on each 10th deploy or when platform is updated. Admin or even user could force an app image update on next deploy via tsuru-client.

ArxdSilva:
fixes #813

### DIFF
--- a/api/app.go
+++ b/api/app.go
@@ -422,7 +422,7 @@ func updateApp(w http.ResponseWriter, r *http.Request, t auth.Token) (err error)
 		wantedPerms = append(wantedPerms, permission.PermAppUpdateRouter)
 	}
 	if updateData.UpdatePlatform {
-		wantedPerms = append(wantedPerms, permission.PermPlatformUpdate)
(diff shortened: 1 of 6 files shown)


api/app.go
```diff
@@ -419,6 +424,9 @@ func updateApp(w http.ResponseWriter, r *http.Request, t auth.Token) (err error)
 	if updateData.Router != "" {
 		wantedPerms = append(wantedPerms, permission.PermAppUpdateRouter)
 	}
+	if updateData.UpdatePlatform {
+		wantedPerms = append(wantedPerms, permission.PermPlatformUpdate)
```
andrestc: `PermPlatformUpdate` is actually used by admins to update platforms installed in tsuru (we should not use it here). We can add a new permission, e.g 'PermAppImageReset', to check here.

ArxdSilva: Would not be better `PermAppUpdateImageReset`?

andrestc: yes!

PR #1597, opened 2017-06-13
app-update platform

### LINKED ISSUE #1591: Allow changing app platform on app-update
With the new `app-deploy-rebuild` functionality, we could add support for changing an app platform. 

If the user requests a new platform for his/her app, tsuru would trigger a rebuild using the new platform as base and check if the deploys succeed; if everything is ok, the app's platform is changed to the new one and if the deploy fails, the app keeps the previous platform.

ArxdSilva:
fixes #1591 

After merge of this and #1590 It'll be needed to do some adaptations on master.

### DIFF
--- a/api/app.go
+++ b/api/app.go
@@ -395,6 +395,7 @@ func updateApp(w http.ResponseWriter, r *http.Request, t auth.Token) (err error)
 		Description:    r.FormValue("description"),
 		Router:         r.FormValue("router"),
 		Tags:           r.Form["tag"],
+		Platform:       r.FormValue("platform"),
 		UpdatePlatform: imageReset,
 	}
 	appName := r.URL.Query().Get(":appname")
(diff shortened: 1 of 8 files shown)


app/app.go
```diff
@@ -349,6 +349,8 @@ func (app *App) Update(updateData App, w io.Writer) (err error) {
 	poolName := updateData.Pool
 	teamOwner := updateData.TeamOwner
 	routerName := updateData.Router
+	platform := updateData.Platform
+	updatePlatform := updateData.UpdatePlatform
```
ggarnier: Do you really need this `updatePlatform` parameter? Couldn't you just set it to true when `plaform` param is set?

ArxdSilva: Done! (:

api/app.go
```diff
@@ -419,6 +420,10 @@ func updateApp(w http.ResponseWriter, r *http.Request, t auth.Token) (err error)
 	if updateData.Router != "" {
 		wantedPerms = append(wantedPerms, permission.PermAppUpdateRouter)
 	}
+	if updateData.Platform != "" {
+		wantedPerms = append(wantedPerms, permission.PermAppUpdatePlatform)
+		updateData.UpdatePlatform = true
+	}
 	if len(wantedPerms) == 0 {
 		msg := "Neither the description, plan, pool, router or team owner were set. You must define at least one."
```
andrestc: should add "or platform" to this error message.

app/app.go
```diff
@@ -391,6 +392,16 @@ func (app *App) Update(updateData App, w io.Writer) (err error) {
 	if tags != nil {
 		app.Tags = tags
 	}
+	if platform != "" {
+		p, errPlat := GetPlatform(platform)
+		if errPlat != nil {
+			return errPlat
+		}
+		app.Platform = p.Name
```
andrestc: I think we could check here if app.Platform is different then p.Name and only if thats true, set app.UpdatePlatform to true.

PR #1600, opened 2017-06-21
App create without platform

### LINKED ISSUE #1491: "platform" should not be mandatory in app-create
In apps that uses  Docker image deployment, the `plataform` information should not be mandatory in app-create.

tsuru should:

* Allow app-create without plataform info 
* Support to change and add platforms using `app-update`
* To track what image/platform is used in a deploy, tsuru should add app platform info in the deploy event data.

ArxdSilva:
fixes #1491

### DIFF
--- a/api/deploy.go
+++ b/api/deploy.go
@@ -88,9 +88,6 @@ func deploy(w http.ResponseWriter, r *http.Request, t auth.Token) (err error) {
 	if err != nil {
 		return &tsuruErrors.HTTP{Code: http.StatusNotFound, Message: err.Error()}
 	}
-	if (instance.GetPlatform() == "") && (origin != "image") {
-		return &tsuruErrors.HTTP{Code: http.StatusBadRequest, Message: "can't deploy app without platform, if it's not an image"}
-	}
 	var build bool
 	buildString := r.FormValue("build")
 	if buildString != "" {
--- a/api/deploy.go
+++ b/api/deploy.go
@@ -89,7 +89,7 @@ func deploy(w http.ResponseWriter, r *http.Request, t auth.Token) (err error) {
 		return &tsuruErrors.HTTP{Code: http.StatusNotFound, Message: err.Error()}
 	}
 	if (instance.GetPlatform() == "") && (origin != "image") {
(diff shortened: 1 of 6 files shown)


api/deploy.go
```diff
@@ -88,6 +88,9 @@ func deploy(w http.ResponseWriter, r *http.Request, t auth.Token) (err error) {
 	if err != nil {
 		return &tsuruErrors.HTTP{Code: http.StatusNotFound, Message: err.Error()}
 	}
+	if (instance.GetPlatform() == "") && (origin != "image") {
+		return &tsuruErrors.HTTP{Code: http.StatusUnauthorized, Message: "can't deploy app without platform, if It's not a docker image"}
```
ggarnier: Just a tiny detail: I would remove `docker` from this error message. And downcase the word `It's`

api/deploy.go
```diff
@@ -88,6 +88,9 @@ func deploy(w http.ResponseWriter, r *http.Request, t auth.Token) (err error) {
 	if err != nil {
 		return &tsuruErrors.HTTP{Code: http.StatusNotFound, Message: err.Error()}
 	}
+	if (instance.GetPlatform() == "") && (origin != "image") {
+		return &tsuruErrors.HTTP{Code: http.StatusUnauthorized, Message: "can't deploy app without platform, if it's not an image"}
```
andrestc: I think an http.StatusBadRequest would fit better here. Also, perhaps we should allow `rollback` too (cc @cezarsa)? Maybe we could move this logic to the `app.ValidateOrigin` func.

cezarsa: I think this logic should be based on `DeployKind` instead of origin. The origin is only informative and can be spoofed.

app/deploy.go
```diff
@@ -252,6 +252,9 @@ func deployToProvisioner(opts *DeployOptions, evt *event.Event) (string, error)
 	if opts.Kind == "" {
 		opts.GetKind()
 	}
+	if ((opts.App.GetPlatform() == "") && (opts.Kind != DeployImage)) || ((opts.App.GetPlatform() == "") && (opts.Kind != DeployRollback)) {
```
andrestc: I think you can write this as:
`((opts.App.GetPlatform() == "") && (opts.Kind != DeployImage && opts.Kind != DeployRollback))`

Issue #1347, opened 2016-01-27
unit-add should start stopped units


ggarnier:
When I stop an app and then add a unit to it, the old units stay stopped, while the new one is started. This lets the app in an inconsistent state (some requests hit the new unit, but others get HTTP 503 when trying to reach one of the stopped units).

Example:

```
$ tsuru app-info -a test
...
Units [web]: 1
+--------------+---------+-----------+-------+------------+
| Unit         | State   | Host      | Port  | IP         |
+--------------+---------+-----------+-------+------------+
| ad0c9727a7db | stopped | 127.0.0.1 | 32799 | 172.17.0.4 |
+--------------+---------+-----------+-------+------------+

$ tsuru unit-add 1 -a test

---- Starting 1 new unit [web: 1] ----
 ---> Started unit be4a3bb684 [web]

---- Binding and checking 1 new unit ----
 ---> Bound and checked unit be4a3bb684 [web]

---- Adding routes to new units ----
 ---> Added route to unit be4a3bb684 [web]

$ tsuru app-info -a test
...
Units [web]: 2
+--------------+----------+-----------+-------+------------+
| Unit         | State    | Host      | Port  | IP         |
+--------------+----------+-----------+-------+------------+
| ad0c9727a7db | stopped  | 127.0.0.1 | 32799 | 172.17.0.4 |
| be4a3bb68474 | starting | 127.0.0.1 | 32800 | 172.17.0.4 |
+--------------+----------+-----------+-------+------------+
```

tarsisazevedo:
I think tsuru should disallow to add new units in stopped apps, what do you think?

fsouza:
Or maybe add units in stopped state?

ggarnier:
I see 3 possible solutions here:

1 - disallow to add new units when current units aren't started (stopped or asleep, for example)
2 - add new units with the same state of the old units
3 - start all units after adding

I think the third one makes more sense

cezarsa:
Adding units without starting them would cause their healthcheck and startup hooks to never be executed. Possible errors would only show up later when the user tries to start the units. On a normal workflow these errors would happen during the unit-add command making the command abort and rollback the added units.

I'm voting for either 1 or 3. The most explicity solution seems to be 1 and if the user wants the behavior of 3 it's as simple as `tsuru app-start && tsuru unit-add`.

tarsisazevedo:
I agree @cezarsa, but we can mix two options, 1 and 3, in one feature.
[...]

PR #1602, opened 2017-06-21
api: prevents adding units in stopped apps

ArxdSilva:
fixes tsuru/tsuru#1347

### DIFF
--- a/api/app.go
+++ b/api/app.go
@@ -501,6 +501,22 @@ func addUnits(w http.ResponseWriter, r *http.Request, t auth.Token) (err error) 
 	if !allowed {
 		return permission.ErrUnauthorized
 	}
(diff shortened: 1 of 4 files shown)


andrestc:
Should we allow adding units *only* if all current units are started or should we allow adding units if there are some units in `starting`, `error` etc states? ping @cezarsa @ggarnier

cezarsa:
Yeah, I think we should only prevent the addition of new units if they are with StatusStopped. Also, this logic should move to the `app` package.

ArxdSilva:
@cezarsa that was something that kept me thinking about where it'd be best, thanks!

ggarnier:
@cezarsa StatusStopped or StatusAsleep

Issue #1475, opened 2016-10-05
Kubernetes provisioner


andrewsmedina:
Now that tsuru supports more than one provisioner, we can add a provisioner based on Kubernetes.

Kubernetes is a production ready container orchestration used by a lot of companies: http://kubernetes.io/case-studies/

heynemann:
AWE-SO-ME!

bgrant0607:
Let us know how we can help.

andrewsmedina:
Hi @bgrant0607 do you want help with code?

bgrant0607:
@andrewsmedina I was more thinking about answering questions, giving you pointers to code/docs, helping you understand the K8s roadmap if there's something you need that doesn't exist yet, etc.

PR #1603, opened 2017-06-22
Kubernetes integration test

ggarnier:
Now it's possible to set multiple cluster providers in the integration test, with the environment variable `TSURU_INTEGRATION_clusters`.

The current available cluster providers are GCE and minikube.


ggarnier:
Related to #1475

Issue #1589, opened 2017-06-06
Unable to unset private variable


bcdonadio:
When you set a private variable, you're unable to unset it until you overwrite it with a public variable and then try to unset it again.

Steps to reproduce:
```bash
tsuru env-set -a app test1=x
tsuru env-get -a app #test1 exists as expected
tsuru env-unset -a app test1
tsuru env-get -a app #test1 doesn't exists as expected

tsuru env-set -a app -p test2=x
tsuru env-get -a app #test2 exists as expected
tsuru env-unset -a app test2
tsuru env-get -a app #test2 STILL EXISTS, unexpectedly

tsuru env-set -a app test2=x
tsuru env-unset -a app test2
tsuru env-get -a app #now test2 is properly removed
```

Using tsurud 1.2.2.

andrestc:
I was able to reproduce this on 1.3.0-rc3. Thanks for reporting.

cezarsa:
Fixed by tsuru/tsuru#1625.

PR #1614, opened 2017-07-07
Fix for #1589

daniellima:
Should fix bug #1589. 

I just wanted to point out that 7408d0e5 explicitly created this behavior but, looking at the code, I found no motive for this. For example, the PublicOnly flag always received the value false everywhere except in a test that verifies that it works. Since this was marked as a bug, I felt free to remove this behavior.

### DIFF
--- a/api/app.go
+++ b/api/app.go
@@ -983,7 +983,7 @@ func unsetEnv(w http.ResponseWriter, r *http.Request, t auth.Token) (err error) 
 	return a.UnsetEnvs(
 		bind.UnsetEnvApp{
 			VariableNames: variables,
-			PublicOnly:    true,
+			PublicOnly:    false,
 			ShouldRestart: !noRestart,
 		}, writer,
 	)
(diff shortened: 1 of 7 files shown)


cezarsa:
So, there are some problems here, this env vars handling code is old and kinda convoluted right now. Private env vars were originally created to be used by envs set by services, this is why there was a commit forbidding the deletion of private env vars in the past. The correct way to remove them would be unbinding the service.

Later on though we decided to allow users to directly set env vars as private, however, this wasn't implemented correctly causing this problem we have being unable to unset them. I agree that we should allow unsetting them but we also must forbid unsetting envs that were injected when binding a service, otherwise this is just too dangerous. Also, new test cases should be added to validate such scenarios.

I can give it a shot on making this look correct and update this PR if you do not mind.

daniellima:
Ok, feel free to do it. I am new to tsuru and Go in general and don't think I will have time to fix this correctly, at least for now.

cezarsa:
This was fixed as a side-effect of tsuru/tsuru#1625.

Issue #1595, opened 2017-06-13
user-list with context returns caller's info


ArxdSilva:
When calling `user-list` with a wrong context value It should return some error in case of empty result, instead of returning the caller's own info.

```shell 
➜  ~ tsuru user list -r team-member -c cartola
+-----------------------------+---------------------+
| User                        | Roles               |
+-----------------------------+---------------------+
| arthur.silva@corp.globo.com | admin(global)       |
|                             | team-create(global) |
+-----------------------------+---------------------+
```

PR #1617, opened 2017-07-11
api: Fixed user list. Added a message for when the context is wrong.

rafaelfragosom:
References #1595


ggarnier:
@rafaelfragosom thanks!

PR #1627, opened 2017-07-27
event: blocks children events when parent is blocked

andrestc:
With this commit if the users block e.g "app.update"
any event that starts with "app.update" will be blocked,
e.g, "app.update.add.unit"

### DIFF
--- a/event/block.go
+++ b/event/block.go
@@ -49,6 +49,19 @@ type Block struct {
 	Active    bool
 }
 
+func (b *Block) Blocks(e *Event) bool {
+	if !(strings.HasPrefix(e.Kind.Name, b.KindName) || b.KindName == "") {
+		return false
+	}
+	if !(e.Owner.Name == b.OwnerName || b.OwnerName == "") {
+		return false
+	}
+	if !(e.Target == b.Target || b.Target == Target{} || (b.Target.Type == e.Target.Type && b.Target.Value == "")) {
+		return false
+	}
+	return true
+}
+
 func (b *Block) String() string {
 	kind := b.KindName
 	if kind == "" {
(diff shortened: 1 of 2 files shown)


event/block.go
```diff
@@ -116,15 +117,21 @@ func checkIsBlocked(evt *Event) error {
...
 		{"active": true},
-		{"$or": []bson.M{{"kindname": evt.Kind.Name}, {"kindname": ""}}},
+		{"$or": kindQuery},
 		{"$or": []bson.M{{"ownername": evt.Owner.Name}, {"ownername": ""}}},
 		{"$or": []bson.M{
 			{"target": evt.Target},
 			{"target": bson.M{"$exists": false}},
 			{"target.type": evt.Target.Type, "target.value": ""}}},
 	}}
+	fmt.Printf("%#+v\n", query)
```
cezarsa: debug?

event/block.go
```diff
@@ -116,21 +117,21 @@ func checkIsBlocked(evt *Event) error {
...
-	if len(blocks) > 0 {
-		return &ErrEventBlocked{event: evt, block: &blocks[0]}
+	for _, b := range blocks {
+		if !(strings.HasPrefix(evt.Kind.Name, b.KindName) || b.KindName == "") {
+			continue
+		}
+		if !(evt.Owner.Name == b.OwnerName || b.OwnerName == "") {
+			continue
+		}
+		if !(evt.Target == b.Target || b.Target == Target{} || b.Target.Type == evt.Target.Type && b.Target.Value == "") {
```
cezarsa: Maybe we should add some extra parenthesis on the last condition just to make precedence cleared when reading? `evt.Target == b.Target || b.Target == Target{} || (b.Target.Type == evt.Target.Type && b.Target.Value == "")`

event/block.go
```diff
@@ -116,21 +117,21 @@ func checkIsBlocked(evt *Event) error {
...
+	for _, b := range blocks {
+		if !(strings.HasPrefix(evt.Kind.Name, b.KindName) || b.KindName == "") {
+			continue
+		}
+		if !(evt.Owner.Name == b.OwnerName || b.OwnerName == "") {
+			continue
+		}
+		if !(evt.Target == b.Target || b.Target == Target{} || (b.Target.Type == evt.Target.Type && b.Target.Value == "")) {
+			continue
+		}
```
ggarnier: I think this code would be easier to understand if you extracted these 3 `if`s to a function, like `blockMatchesEvent(b, evt)`

PR #1633, opened 2017-07-31
api,all: refactor server shutdown logic

andrestc:
This PR refactors some of our shutdown logic and removes the need of the graceful.Server.

Since go 1.8, the http.Server has a Shutdown function that can be called to safely close idle connections before stopping the server, removing the need of the tylerb/graceful server.

Also, Adding a context and an error return to the Shutdown func makes it possible to register the http.Server as a Shutdownable. Our Shutdownable implementations can also start making use of those to properly handle timeout and errors on the Shutdown process.


andrestc:
The shutdown output is now:
```
tsuru is shutting down, waiting for pending connections to finish.
running shutdown for node healer...
running shutdown for log pub/sub connections...
running shutdown for log pub/sub connections. DONE.
running shutdown for *http.Server...
running shutdown for queued tasks...
running shutdown for node healer. DONE.
running shutdown for *http.Server. DONE.
Listening stopped: http: Server closed
running shutdown for queued tasks. DONE.
```

The `Stringer` thing was added so the output for http.Server would look a bit better.

PR #1634, opened 2017-08-01
api/context: replaces gorilla/context with stdlib context

### LINKED ISSUE #1503: Proposal: use native context instead of gorilla/context
Go 1.7 came with new feature: new way to set objects inside a http.Request

example:

```go
import "context"
func myMiddleware(w http.ResponseWriter, r *http.Request) {
    newCtx := context.WithValue(r.Context(), "app", &App{...})
    newReq := r.WithContext(newCtx)
    nextHandler(w, newReq)
}

func RecoverApp(r *http.Request) *App {
  if app, ok := r.Context().Value("app").(*App); ok {
    return app
  }
  return nil
}

andrestc:
This PR replaces the use of gorilla/context in the api/context pkg with the std library context (introduced in go 1.7).

Fixes #1503

### DIFF
--- a/api/app_test.go
+++ b/api/app_test.go
@@ -4937,7 +4937,9 @@ func (s *S) TestGetApp(c *check.C) {
 	c.Assert(err, check.IsNil)
 	expected, err := app.GetByName(a.Name)
 	c.Assert(err, check.IsNil)
-	app, err := getAppFromContext(a.Name, nil)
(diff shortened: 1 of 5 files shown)


andrestc:
@cezarsa I think we still have some dependencies on `gorilla/context`, namely, the `gandalftest` uses `gorilla/pat` that currently relies on `gorilla/context`. I think we can try doing a PR and seeing if they do accept it but do you think this blocks this PR?

PR #1637, opened 2017-08-02
Team repository

ggarnier:
Extracts all accesses to the teams collection in the DB to a repository.

I still want to extract the `Team` struct to its own package, together with other domain specific rules. With this, we won't need to translate between `auth.Team` and `storage.Team` structs anymore.

Eventually, the `db` package could also be moved to `storage`, to stop exposing direct access to the database.


ggarnier:
Refs #1994

PR #1640, opened 2017-08-03
registry: create package with RemoveAppImages method

guilhermebr:


registry/docker/client.go
```diff
@@ -0,0 +1,54 @@
...
+		req.Header.Set(k, v)
+	}
+	resp, err := r.HTTPClient.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	return resp, nil
+}
+
+func (r *dockerRegistry) client() (err error) {
```
andrestc: maybe use one of the clients in https://github.com/tsuru/tsuru/blob/master/net/client.go#L37?

registry/docker/registry.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+	for _, tag := range tags {
+		digest, err := r.getDigest(image, tag)
+		if err != nil {
+			fmt.Printf("failed to get digest for image %s/%s:%s on registry: %v\n", r.server, image, tag, err)
+			continue
+		}
+		err = r.removeImage(image, digest)
+		if err != nil {
+			if strings.Contains(err.Error(), "405") {
```
ggarnier: It would be nice to set a specific error here (something like `ErrMethodNotAllowed`), instead of parsing the error message.

registry/docker/registry.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func (r dockerRegistry) removeImage(image, digest string) error {
+	path := fmt.Sprintf("/v2/%s/manifests/%s", image, digest)
+	resp, err := r.doRequest("DELETE", path, nil)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode == 404 {
```
ggarnier: You could replace these `404` and `405` with `http.StatusNotFound ` and `http.StatusMethodNotAllowed`

registry/docker/testing/server.go
```diff
@@ -0,0 +1,156 @@
...
+	"encoding/json"
+	"fmt"
+	"net"
+	"net/http"
+	"sync"
+
+	"github.com/gorilla/mux"
+)
+
+type Repository struct {
```
ggarnier: I think we could use a different name here, like `ImageRepository`, to avoid confusion with the `repository` package.

registry/registry.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func (r dockerRegistry) getDigest(image, tag string) (string, error) {
+	path := fmt.Sprintf("/v2/%s/manifests/%s", image, tag)
+	resp, err := r.doRequest("HEAD", path, map[string]string{"Accept": "application/vnd.docker.distribution.manifest.v2+json"})
+	if err != nil {
+		return "", err
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode == 404 {
```
ggarnier: You could replace `404` with `http.StatusNotFound`

registry/registry.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func (r dockerRegistry) getImageTags(image string) ([]string, error) {
+	path := fmt.Sprintf("/v2/%s/tags/list", image)
+	resp, err := r.doRequest("GET", path, nil)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode == 404 {
```
ggarnier: Here too

PR #1644, opened 2017-08-07
all: allow renaming team and updating all references to the old name

cezarsa:


api/auth.go
```diff
@@ -226,6 +232,108 @@ func resetPassword(w http.ResponseWriter, r *http.Request) (err error) {
...
+	})
+	if err != nil {
+		return err
+	}
+	defer func() { evt.Done(err) }()
+	u, err := t.User()
+	if err != nil {
+		return err
+	}
+	defer func() {
```
ggarnier: Shouldn't this `defer` be defined after `CreateTeam`?

cezarsa: Right on, fixed now.

api/auth.go
```diff
@@ -226,6 +232,99 @@ func resetPassword(w http.ResponseWriter, r *http.Request) (err error) {
...
+	volume.RenameTeam,
+	provision.RenamePoolTeam,
+}
+
+// title: team update
+// path: /teams/{name}
+// method: POST
+// consume: application/x-www-form-urlencoded
+// responses:
+//   200: Team created
```
ArxdSilva: shouldnt this be 'team updated'?

cezarsa: fixed!

PR #1646, opened 2017-08-07
Split deploy into build and deploy commands

guilhermebr:


api/app.go
```diff
@@ -1692,7 +1692,11 @@ func registerUnit(w http.ResponseWriter, r *http.Request, t auth.Token) error {
 	}
 	err = a.RegisterUnit(hostname, customData)
 	if err != nil {
-		if _, ok := err.(*provision.UnitNotFoundError); ok {
+		if err, ok := err.(*provision.UnitNotFoundError); ok {
+			if r.Header.Get("x-Agent-Version") == "" {
```
andrestc: maybe capitalize  **X**-Agent-Version?

api/app_test.go
```diff
@@ -5334,12 +5334,37 @@ func (s *S) TestRegisterUnitInvalidUnit(c *check.C) {
 	request, err := http.NewRequest("POST", "/apps/myappx/units/register", body)
 	c.Assert(err, check.IsNil)
 	request.Header.Set("Authorization", "bearer "+s.token.GetValue())
+	request.Header.Set("x-Agent-Version", "0.2.4")
```
andrestc: same as above

api/app.go
```diff
@@ -1692,7 +1692,11 @@ func registerUnit(w http.ResponseWriter, r *http.Request, t auth.Token) error {
...
+				msgError := fmt.Sprintf("Please contact admin. %s platform is using outdated deploy-agent version, minimum required version is 0.2.4", a.GetPlatform())
+				return &errors.HTTP{Code: http.StatusUpgradeRequired, Message: msgError}
```
andrestc: i would just return `http.StatusBadRequest` here as `http.StatusUpgradeRequired` actually means that the protocol version is not supported

andrestc:
LGTM overall expect for a few comments

PR #1654, opened 2017-08-10
all: allow restricting services to certain pools

### LINKED ISSUE #1587: Allow restricting services to specific pools
Currently, we can constraint which teams can use a pool and which routers are available for each pool. It would be good if one could constraint which services are available for each pool.

ArxdSilva:
fixes #1587

### DIFF
--- a/provision/pool/pool.go
+++ b/provision/pool/pool.go
@@ -2,19 +2,19 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-package provision
+package pool
 
 import (
 	"encoding/json"
 	"fmt"
 	"regexp"
-	"sort"
 	"strings"
 
 	"github.com/pkg/errors"
 	"github.com/tsuru/tsuru/auth"
 	"github.com/tsuru/tsuru/db"
 	tsuruErrors "github.com/tsuru/tsuru/errors"
+	"github.com/tsuru/tsuru/provision"
 	"github.com/tsuru/tsuru/router"
 	"github.com/tsuru/tsuru/service"
 	"github.com/tsuru/tsuru/validation"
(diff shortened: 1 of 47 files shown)


provision/pool/pool.go
```diff
@@ -119,6 +132,20 @@ func (p *Pool) GetDefaultRouter() (string, error) {
 	return router.Default()
 }
 
+// var ValRegistry *valuesRegistry
```
cezarsa: Please remove this commented code

service/service.go
```diff
@@ -32,6 +32,10 @@ var (
 	ErrServiceAlreadyExists = errors.New("Service already exists.")
 )
 
+// func init() {
```
cezarsa: This block should also be removed.

service/service.go
```diff
@@ -239,3 +243,15 @@ func RenameServiceTeam(oldName, newName string) error {
 	_, err = bulk.Run()
 	return err
 }
+
+func servicesNames() []string {
```
cezarsa: Is this function unused? If so, please remove it.

provision/pool/pool_test.go
```diff
@@ -39,17 +42,6 @@ func (s *S) TearDownSuite(c *check.C) {
 	s.storage.Close()
 }
 
-func (s *S) SetUpTest(c *check.C) {
```
cezarsa: SetUpTest should continue to exist with a call to `dbtest.ClearAllCollections(s.storage.Apps().Database)` to ensure a clean db on each test.
Also, the database:name in `SetUpSuite` should be changed not to conflict with the database:name used in `provision/suite_test.go`, because tests from different packages run in parallel.

ArxdSilva: Done, forgot that code from import cycle problems, also changed the rest to meet what was requested :D

cezarsa:
Just fix the variable shadowing in `app/app.go:907` and I think we're ready to go.

ArxdSilva:
Thanks @cezarsa 👍

PR #1658, opened 2017-08-11
provision/swarm: use cluster pkg in swarm provisioner

cezarsa:


api/cluster.go
```diff
@@ -27,6 +47,65 @@ import (
...
+	if err == nil {
+		err = dec.DecodeValues(&provCluster, r.Form)
+	}
+	if err != nil {
+		return &tsuruErrors.HTTP{
+			Code:    http.StatusBadRequest,
+			Message: err.Error(),
+		}
+	}
+	err = createClusterMachine(&provCluster)
```
ggarnier: Shouldn't this be called only after you check that there isn't already a cluster with that name (line 84)?

cezarsa: Yep, fixed.

PR #1660, opened 2017-08-11
autoscale: always enabled and skipping provisioners with error

### LINKED ISSUE #1456: autoscale should be active besides docker:auto-scale:enabled
tsuru supports autoscale creation at runtime, but if it's not previously enabled in conf file, it will never run.

andrestc:
This PR removes the need to enable the autoscale using the configuration file and only relies on the rules registered for each pool.

This also makes sure we skip provisioners with error instead of checking other provisioners nodes.

Fixes #1456.


andrestc:
PTAL @cezarsa @ggarnier

PR #1665, opened 2017-08-15
Enhancement/implements resource team info

### LINKED ISSUE #1610: team-info
Now If you need to get the users of a team, It's needed to get some infos first, such as: roles in that team and team name. The only way of doing this is by doing to each role the following:
```shell
➜ ~ tsuru user-list -r <role> -c <team>
+-----------------------------+---------------------+
| User                        | Roles               |
+-----------------------------+---------------------+
| garen@corp.globo.com        | demacia(global)     |
|                             | team-member(global) |
+-----------------------------+---------------------+
```

It would be nice to have some command that outputs the same info and much more about that team, such as: users, pools, services, apps, quota. It would be something like:
```shell
➜ ~ tsuru team-info <team>
```

guiferpa:
fix #1610


api/server.go
```diff
@@ -221,6 +221,7 @@ func RunServer(dry bool) http.Handler {
 	m.Add("1.0", "Post", "/teams", AuthorizationRequiredHandler(createTeam))
 	m.Add("1.0", "Delete", "/teams/{name}", AuthorizationRequiredHandler(removeTeam))
 	m.Add("1.4", "Post", "/teams/{name}", AuthorizationRequiredHandler(updateTeam))
+	m.Add("1.0", "Get", "/teams/{name}", AuthorizationRequiredHandler(teamInfo))
```
andrestc: this should be "1.4" instead of "1.0" as the route is being added on tsuru v1.4.

api/auth.go
```diff
@@ -461,6 +461,85 @@ func teamList(w http.ResponseWriter, r *http.Request, t auth.Token) error {
...
+	team, err := auth.GetTeam(teamName)
+	if err != nil {
+		return err
+	}
+
+	_, err = t.Permissions()
+	if err != nil {
+		return err
+	}
+
```
andrestc: please, remove the blank lines between statements

api/auth.go
```diff
@@ -461,6 +461,85 @@ func teamList(w http.ResponseWriter, r *http.Request, t auth.Token) error {
...
+	apps, err := app.List(&app.Filter{})
```
andrestc: You can pass down the Team as a filter to app.List() so you don't need to loop over the returned apps
something as: `app.List(&app.Filter{Extra: map[string][]string{"teams": team.Name}})`

api/auth.go
```diff
@@ -461,6 +461,85 @@ func teamList(w http.ResponseWriter, r *http.Request, t auth.Token) error {
...
+	for _, app := range apps {
+		for _, appTeam := range app.Teams {
+			if appTeam == team.Name {
+				includedApps = append(includedApps, app)
+				break
+			}
+		}
+	}
+
+	pools, err := provision.ListAllPools()
```
andrestc: You could use `provision.ListPoolsForTeam`

api/auth.go
```diff
@@ -461,6 +461,85 @@ func teamList(w http.ResponseWriter, r *http.Request, t auth.Token) error {
...
+//   200: Info team
+//   404: Not found
+//   401: Unauthorized
+func teamInfo(w http.ResponseWriter, r *http.Request, t auth.Token) error {
+	teamName := r.URL.Query().Get(":name")
+	team, err := auth.GetTeam(teamName)
+	if err != nil {
+		return err
+	}
+
```
andrestc: Before doing anything, I think you should probably check if the user token `t` has permission to read the team information, using `PermTeamRead`. You can see how other handlers do that, e.g,  https://github.com/tsuru/tsuru/blob/master/api/app.go#L230.

api/auth.go
andrestc: We should probably have a different list of apps that include only the ones owned by the team. You can use the `TeamOwner` field on the app.Filter.

api/auth.go
```diff
@@ -461,6 +461,85 @@ func teamList(w http.ResponseWriter, r *http.Request, t auth.Token) error {
...
+				perms, _ := t.Permissions()
+				userData, _ := createAPIUser(perms, &user, roleMap, includeAll)
+				includedUsers = append(includedUsers, userData)
+				break
+			}
+		}
+	}
+
+	var result map[string]interface{}
+	result = map[string]interface{}{
```
andrestc: This should be a `map[string][]string` and all fields should just include the name, e.g, instead of `users` being the apiUser struct, should just include the user names

api/auth.go
guiferpa: Currently the package to `.ListPoolsForTeam` is `pool`, right?

api/auth.go
```diff
@@ -461,6 +461,85 @@ func teamList(w http.ResponseWriter, r *http.Request, t auth.Token) error {
...
+	}
+
+	users, err := auth.ListUsers()
+	if err != nil {
+		return err
+	}
+	includedUsers := make([]*apiUser, 0)
+	for _, user := range users {
+		for _, role := range user.Roles {
+			if role.ContextValue == team.Name {
```
cezarsa: I don't think that's how we should get users for a team. The meaning here should be users with roles with ContextType == CtxTeam and a ContextValue equal to the current team. See my suggestion below for the output grouped by role.
I think this loop would look something like: (untested pseudocode)
```
roleUserMap := map[string][]string{}
for _, user := range users {
  for _, roleInstance := range user.Roles {
    role := permission.FindRole(roleInstance.Name)
    if role.ContextType == permission.CtxTeam && roleInstance.ContextValue == team.Name {
      roleUserMap[role.Name] = append(roleUserMap[role.Name], user.Email)
    }
  }
}
```
You could cache FindRole results in a map to avoid repeated lookups to the same role in different users.

api/auth.go
guiferpa: I gonna create some tests for `.ListPoolsForTeam`, I think something wrong is happening.

andrestc: yes, It would be nice to add tests to the handler also.

guiferpa: With this command the Tsuru API create an `pool constraint` with correct and valid context
- `tsuru pool-constraint-set pool1 team team1`

With this command the Tsuru API create an `pool constraint` with fail and invalid context
- `tsuru pool-constraint-set pool1 slash-gnr team1`

I found this case when used `.ListPoolsForTeam` and not get info about pool, I used `tsuru pool-constraint-set pool1 team1` as command, and the Tsuru API created `team1` as context and not value in database.

I think this case a bug, no there context validate.

What do you think about?

andrestc: What version of tsuru client are you using? I can do that:
```$ tsuru pool constraint set pool1 shalsh-gnr team1
Error: invalid constraint type. Valid types are: team,router,service
```

api/auth.go
guiferpa:
```json
{
    ...
    "Roles": [
        {
            "Name": "AllowAll",
            "ContextType": "global",
            "ContextValue": "team1"
        }
    ]
    ...
}
```
> User response scope from /teams/{name}

So, but by example, I need `Roles` from `apiUser` in my tsuru CLI for format an roles output. What to do this case?

api/auth.go
guiferpa: tsuru version 1.3.0-rc2

guiferpa:
Up

api/auth.go
```diff
@@ -461,6 +461,67 @@ func teamList(w http.ResponseWriter, r *http.Request, t auth.Token) error {
...
+		return err
+	}
+	users, err := auth.ListUsers()
+	if err != nil {
+		return err
+	}
+	includedUsers := make([]*apiUser, 0)
+	for _, user := range users {
+		for _, roleInstance := range user.Roles {
+			role, err := permission.FindRole(roleInstance.Name)
```
ggarnier: You could cache the result of this function, to avoid calling it multiple times for the same role

api/auth.go
```diff
@@ -461,6 +461,67 @@ func teamList(w http.ResponseWriter, r *http.Request, t auth.Token) error {
...
 
+// title: team info
+// path: /teams/{name}
+// method: GET
+// produce: application/json
+// responses:
+//   200: Info team
+//   404: Not found
+//   401: Unauthorized
+func teamInfo(w http.ResponseWriter, r *http.Request, t auth.Token) error {
```
ggarnier: Could you add some tests for this?

api/auth.go
```diff
@@ -461,6 +461,67 @@ func teamList(w http.ResponseWriter, r *http.Request, t auth.Token) error {
...
+		for _, roleInstance := range user.Roles {
+			role, err := permission.FindRole(roleInstance.Name)
+			if err != nil {
+				break
+			}
+			if role.ContextType == permission.CtxGlobal || (role.ContextType == permission.CtxTeam && roleInstance.ContextValue == team.Name) {
+				canInclude := permission.Check(t, permission.PermTeam)
+				if canInclude {
+					roleMap := make(map[string]*permission.Role)
+					perms, _ := t.Permissions()
```
ArxdSilva: @guiferpa It's best not to ignore these errors

api/auth.go
guiferpa: I do not understand, Do I need sweep the user.Roles for get roleInstance.Name ?

ggarnier: You are calling `permission.FindRole` for each role of each user. If a team has 10 different users with a common role, you would be making the same request 10 times.

A suggestion to improve this would be creating a map with the role names as keys, and the roles as values. Before calling `permission.FindRole`, you could check if that value is already in the map to avoid these unnecessary requests.

Here's an example of how you could do it:

```
cachedRoles := make(map[string]Role)
for _, user := range users {
	for _, roleInstance := range user.Roles {
		role, ok := cachedRoles[roleInstance.Name]
		if !ok {
			role, err = permission.FindRole(roleInstance.Name)
			if err != nil {...}
			cachedRoles[roleInstance.Name] = role
		}
		...
	}
}
```

api/auth.go
```diff
@@ -461,6 +461,78 @@ func teamList(w http.ResponseWriter, r *http.Request, t auth.Token) error {
...
+				}
+				cachedRoles[roleInstance.Name] = role
+			}
+			if role.ContextType == permission.CtxGlobal || (role.ContextType == permission.CtxTeam && roleInstance.ContextValue == team.Name) {
+				canInclude := permission.Check(t, permission.PermTeam)
+				if canInclude {
+					roleMap := make(map[string]*permission.Role)
+					perms, err := t.Permissions()
+					if err != nil {
+						break
```
ArxdSilva: Here you can return the error so the user can handle them.

guiferpa: Like this?
```golang
...
for _, roleInstance := range user.Roles {
  role, ok := cachedRoles[roleInstance.Name]
  if !ok {
   role, err := permission.FindRole(roleInstance.Name)
   if err != nil {
     return &errors.HTTP{Code: http.StatusInternalServerError, Message: err.Error()}
   }
   cachedRoles[roleInstance.Name] = role
  }
...
```

ArxdSilva: yes, this way

cezarsa:
Thank you for your contribution!

PR #1707, opened 2017-08-30
docs: add required migration to 1.4.0 release notes

cezarsa:


andrestc:
We should probably mention that platforms need to be updated

cezarsa:
Done

PR #1736, opened 2017-09-18
add support for multiple routers in one app

cezarsa:


api/server.go
```diff
@@ -167,6 +167,10 @@ func RunServer(dry bool) http.Handler {
 	m.Add("1.2", "Put", "/apps/{app}/certificate", AuthorizationRequiredHandler(setCertificate))
 	m.Add("1.2", "Delete", "/apps/{app}/certificate", AuthorizationRequiredHandler(unsetCertificate))
 
+	m.Add("1.4", "Post", "/apps/{app}/routers", AuthorizationRequiredHandler(addAppRouter))
```
guilhermebr: It's not "1.5"?

cezarsa: Oops, yeah, changing to 1.5.

PR #1737, opened 2017-09-18
call SetUpdatePlatform in app.Deploy for Image and Rollback

guilhermebr:


provision/kubernetes/provisioner.go
```diff
@@ -1,4 +1,3 @@
-// Copyright 2016 tsuru authors. All rights reserved.
```
cezarsa: Removed by accident?

Issue #1740, opened 2017-09-21
Support multiple kind filters in event list


ggarnier:
It would be nice if we could filter the event list by multiple kinds, like `app.deploy` and `app.update`.

PR #1742, opened 2017-09-21
Events filter by multiple kinds

ggarnier:
Allows multiple `kindname` params in the events api.

Refs #1740


event/event.go
```diff
@@ -389,7 +389,7 @@ type TargetFilter struct {
 type Filter struct {
 	Target         Target
 	KindType       kindType
-	KindName       string
+	KindName       []string `form:"-"`
```
andrestc: just a nitpick: maybe we should call this KindName**s**?

ggarnier: Agreed!

PR #1765, opened 2017-10-04
Make rebalance event cancelable

cezarsa:
~Depends on merging https://github.com/tsuru/docker-cluster/pull/35 and adding it here.~


event/event.go
```diff
@@ -939,6 +944,32 @@ func (e *Event) Write(data []byte) (int, error) {
 	return e.logBuffer.Write(data)
 }
 
+func (e *Event) CancelableContext(ctx context.Context) (context.Context, context.CancelFunc) {
+	ctx, cancel := context.WithCancel(ctx)
+	go func() {
+		for {
+			canceled, err := e.AckCancel()
+			if err != nil {
+				if err == ErrNotCancelable {
```
andrestc: Is this needed? `e.AckCancel` returns `nil` when the event is not cancelable: https://github.com/tsuru/tsuru/blob/71a5faae63ac65f0263578e08a5722349aa94b1d/event/event.go#L1006

cezarsa: You're right, I'll move the this check out of here to avoid starting this goroutine if the event is not cancelable.

Issue #1777, opened 2017-10-09
client: cluster-remove doesn't ask for confirmation


ggarnier:
`cluster-remove` command has a `-y/--assume-yes` flag, but even if it's omitted, the command doesn't ask for confirmation before removing the cluster.

anchepiece:
Looks like the documentation is missing as well:
http://tsuru-client.readthedocs.io/en/master/reference.html#cluster-management

I'll take a quick poke around.

ggarnier:
@anchepiece great! Let us know if you need some help

anchepiece:
Created a pull request here:
https://github.com/tsuru/tsuru-client/pull/65

I'll check through the command listing for similar missing checks.
Hope that helps!

anchepiece:
@ggarnier Just a heads up, the documentation was updated and correctly reflects the client arguments (I like to follow through with any commits)

PR #1778, opened 2017-10-09
client: Update install test to account for cluster-remove confirmation #1777

anchepiece:
cluster-remove will ask for confirmation to proceed. This test has been updated to account for the change.

https://github.com/tsuru/tsuru/issues/1777


ggarnier:
@anchepiece thanks!

Issue #1696, opened 2017-08-25
Cluster pool validation


ggarnier:
`cluster-add` command doesn't validate if the pool exists

VitorFalcao:
@ggarnier I would like to fix this bug. Do you have any tips or preference on how it should be fixed?

ggarnier:
@VitorFalcao I think the easiest way is checking how we validate input data in other routes, like team creation:
https://github.com/tsuru/tsuru/blob/afc0745897538d611768016338c393118af2bd72/auth/team.go#L54

Please let me now if you need some help

VitorFalcao:
Thank you. If the pool does not exist, should the function return an "errors.New()"? If so, what should be the message?

ggarnier:
You can do something like this:

https://github.com/tsuru/tsuru/blob/f9d62e093a53bb10e3e82b96de5553fc8535f7e9/api/pool.go#L183-L186

VitorFalcao:
Should I write a new test?

ggarnier:
I think the best option is the third one. You can check if each pool exists before saving the cluster, and only save if all pools exist.

PR #1783, opened 2017-10-16
Validate if pool exists before adding a cluster to it

VitorFalcao:
Fix issue #1696


VitorFalcao:
@andrestc Sure! Working on it.

VitorFalcao:
@andrestc Should I also validate if the pool exists in the method `updateCluster` in `api/cluster`? And before the `createCluster` method there is a comment section specifying the possible return codes and their meanings. Should I add 404 to that list?

andrestc:
@VitorFalcao yes, i think so.

VitorFalcao:
@andrestc The `updateCluster` method of `api/cluster` already has a 404 code specified for when the cluster does not exist. Is there any problem in returning the same code for when a pool does not exist?

andrestc:
I think we could return 400 in this case.

Em 16 de out de 2017 19:23, "Nope" <notifications@github.com> escreveu:

VitorFalcao:
@andrestc The method already uses that code for `Invalid data`

andrestc:
Thats ok, just make sure to add a message describing the error cause in the
body of the response

Em 16 de out de 2017 19:39, "Nope" <notifications@github.com> escreveu:

ggarnier:
@VitorFalcao thank you!

Issue #1782, opened 2017-10-11
Error message in app creation should include public pools


ggarnier:
When I try to create an app and forget the `--pool` flag, the error message shows a list of pools I can choose. Currently, this message only shows pools my team has access to, but it should also show public pools.

The error message is:

```
Error: you have access to "pool1","pool2" pools. Please choose one in app creation
```

rodrigo-brito:
I will get it.

cezarsa:
Closed by #1786

PR #1786, opened 2017-10-16
Include public pools in app creation error

rodrigo-brito:
I tried to list the public constraints, but in some test, the pool `pool1` are showing with the team's pools. But it is not initialized in the test scope, maybe it has a global behaviour, but I don't know.

Any suggestion?

Ref https://github.com/tsuru/tsuru/issues/1782

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -1001,8 +1001,12 @@ func (app *App) getPoolForApp(poolName string) (string, error) {
 			return "", err
 		}
 		if len(pools) > 1 {
+			publicPools, err := pool.ListPublicPools()
+			if err != nil {
+				return "", err
+			}
 			var names []string
-			for _, p := range pools {
+			for _, p := range append(pools, publicPools...) {
 				names = append(names, fmt.Sprintf("%q", p.Name))
 			}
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -5,3 +5,4 @@ docs/_build
 build
 data
 dist/
+.idea/
```
cezarsa: Could you add this to your [global .gitignore file](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) instead of here? We try not to have IDE specific patterns here.

rodrigo-brito: Ok, no problem!

rodrigo-brito:
@cezarsa the constraint  `"*"` for the field `team` returns the default pool too. Is it correct for this case?

cezarsa:
If you call `getPoolsSatisfyConstraints(false, "team", team)` (notice false as the first argument), you'll have a list of all pools available for the team, both public and private. You can use the return from this call in the error message directly, there's no need to call it with "*" as an argument.

rodrigo-brito:
Fixed :recycle:

rodrigo-brito:
@cezarsa fixed ♻️

cezarsa:
Thanks a lot for contributing!

rodrigo-brito:
😄

Issue #1772, opened 2017-10-06
Moving apps between pools belonging to different provisioners should cleanup old artifacts.


cezarsa:
Moving apps between pools belonging to different provisioners should cleanup everything created by the old provisioner.

VitorFalcao:
Hi @cezarsa I've just started contributing to the project (currently fixing issue #1696) and this issue seems interesting. Could you please elaborate on it?

cezarsa:
Currently we support multiple provisioners, which are responsible for scheduling app units (see https://docs.tsuru.io/master/managing/provisioners.html)
Every pool must have an associated provisioner and every app must belong to pool, this way tsuru knows which provisioner to use for the app.

The bug described in this issue arrises when we use the `tsuru app-update` command to change an app's pool. If the provisioner of the new pool is the same as the one from the old pool everything works fine, on the next app-restart or app-deploy tsuru will create the units on the new pool and the old units will be removed.

However, if the provisioners are different we leave behind all the resources created by the old provisioner, this happens because on the new provisioner it'll be as if the app didn't exist before.

As an example, imagine you have an application `app1` created originally on a pool with provisioner `swarm`. When a deploy happens tsuru will create a new Swarm Service called `app1-web` which will then create containers for this service.
If we then update the application's pool to one with provisioner `kubernetes`, on the next deploy tsuru will simply create a new Kubernetes Deployment and leave the old Swarm Service behind. This happens because the `kubernetes` provisioner is not aware (nor should be) of the existence of a Swarm Service for this application.

One possibly solution would be simply forbidding the pool change if the new pool has a different provisioner.
Another, more complex, solution would involve trying to destroy the resources from the old provisioner once it's been created on the new one. This would involve some changes in the way `app-update` works though, as currently it does not restart the application to make the pool change effective.

VitorFalcao:
@cezarsa Thank you for the explanation. I guess the more complex solution would be better, right? At least, IMHO, it seems more user friendly. What do you think?

cezarsa:
Yep, the more complex solution would be better. The simple solution could be used as a stopgap to avoid inconsistencies while we work on getting the complex one right though.

VitorFalcao:
@cezarsa Should I submit a PR with the stopgap and then work on the complex solution?

cezarsa:
Sure, that would be great.

VitorFalcao:
[...]

PR #1794, opened 2017-10-17
Forbidding the pool change if the new pool has a different provisioner

VitorFalcao:
This is the partial fix for #1772  
Could someone please check if I am using the correct HTTP status below?
```go
return &errors.HTTP{Code: http.StatusForbidden, Message: "Moving apps between pools belonging to different provisioners is not alllowed"}
```
I was in doubt between that status and `http.StatusBadRequest`.
I added a test, but I will remove it once I am done with the final solution discussed in #1772 

Should I add the block containing lines 463-473 to the if block in line 437 in `api/app.go`?

### DIFF
--- a/api/app.go
+++ b/api/app.go
@@ -459,6 +459,19 @@ func updateApp(w http.ResponseWriter, r *http.Request, t auth.Token) (err error)
 			return permission.ErrUnauthorized
 		}
 	}
+	if updateData.Pool != "" {
+		currentPool, err := pool.GetPoolByName(a.GetPool())
+		if err != nil {
+			return err
+		}
+		newPool, err := pool.GetPoolByName(updateData.Pool)
+		if err != nil {
+			return err
+		}
+		if currentPool.Provisioner != newPool.Provisioner {
+			return &errors.HTTP{Code: http.StatusForbidden, Message: "Moving apps between pools belonging to different provisioners is not alllowed"}
+		}
+	}
 	evt, err := event.New(&event.Opts{
 		Target:     appTarget(appName),
 		Kind:       permission.PermAppUpdate,
--- a/api/app.go
+++ b/api/app.go
@@ -392,7 +392,7 @@ func createApp(w http.ResponseWriter, r *http.Request, t auth.Token) (err error)
 // produce: application/x-json-stream
 // responses:
 //   200: App updated
-//	 400: Invalid new pool
+//   400: Invalid new pool
 //   401: Unauthorized
 //   404: Not found
 func updateApp(w http.ResponseWriter, r *http.Request, t auth.Token) (err error) {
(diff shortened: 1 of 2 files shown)


api/app.go
```diff
@@ -459,6 +459,19 @@ func updateApp(w http.ResponseWriter, r *http.Request, t auth.Token) (err error)
...
+		currentPool, err := pool.GetPoolByName(a.GetPool())
+		if err != nil {
+			return err
+		}
+		newPool, err := pool.GetPoolByName(updateData.Pool)
+		if err != nil {
+			return err
+		}
+		if currentPool.Provisioner != newPool.Provisioner {
+			return &errors.HTTP{Code: http.StatusForbidden, Message: "Moving apps between pools belonging to different provisioners is not alllowed"}
```
andrestc: I think it would be better to go for `http.StatusBadRequest` as we've been using `http.StatusForbidden` for credentials only.

andrestc:
```
Should I add the block containing lines 463-473 to the if block in line 437 in api/app.go?
```
I think its ok as is

api/app.go
```diff
@@ -459,6 +460,19 @@ func updateApp(w http.ResponseWriter, r *http.Request, t auth.Token) (err error)
...
+		currentPool, err := pool.GetPoolByName(a.GetPool())
+		if err != nil {
+			return err
+		}
+		newPool, err := pool.GetPoolByName(updateData.Pool)
+		if err != nil {
+			return err
+		}
+		if currentPool.Provisioner != newPool.Provisioner {
+			return &errors.HTTP{Code: http.StatusBadRequest, Message: "Moving apps between pools belonging to different provisioners is not alllowed"}
```
cezarsa: Typo: alllowed -> allowed

api/app.go
```diff
@@ -392,6 +392,7 @@ func createApp(w http.ResponseWriter, r *http.Request, t auth.Token) (err error)
 // produce: application/x-json-stream
 // responses:
 //   200: App updated
+//	 400: Invalid new pool
```
cezarsa: Please use only spaces in the comment instead of tab+space

api/app.go
```diff
@@ -459,6 +460,19 @@ func updateApp(w http.ResponseWriter, r *http.Request, t auth.Token) (err error)
 			return permission.ErrUnauthorized
 		}
 	}
+	if updateData.Pool != "" {
+		currentPool, err := pool.GetPoolByName(a.GetPool())
```
cezarsa: Linter failed here with 
```
api/app.go:464::warning: declaration of "err" shadows declaration at api/app.go:398 (vetshadow)
```
To avoid this you can declare `var currentPool, newPool *provision.Pool` in the beginning of the if block, and use `=` instead of `:=`

VitorFalcao: @cezarsa Could you please explain to me why this happens?

VitorFalcao:
Thank you guys for the feedback and sorry for the silly mistakes

api/app.go
cezarsa: Variables in Go have block scope, by using `:=` inside the if block you're actually declaring a new variable `err` that only happens to have the same name as the external `err` variable. This can cause subtle and hard to find errors as changing the value of the inner variable doesn't change the outer one. That's the reason why go vet complains about it. Check https://play.golang.org/p/JD-at0fAbg for a better example.

VitorFalcao: Thank you very much. I wasn't aware of the block scope.

cezarsa:
Thanks a lot for working on this!

PR #1806, opened 2017-10-23
router/api: sends swap data as json

andrestc:


router/api/router.go
```diff
@@ -54,6 +52,11 @@ type routesReq struct {
 	Addresses []string `json:"addresses"`
 }
 
+type swapReq struct {
+	Target    string
```
cezarsa: Other structs here add tags to make json fields lowercase, shouldn't we do the same thing here?

andrestc:
done!


André Carvalho
===============================================
+55 21 98739-1097 / br.linkedin.com/in/andrestc/


2017-10-23 10:45 GMT-02:00 Cezar Sá Espinola <notifications@github.com>:

router/api/router.go
andrestc: done

PR #1816, opened 2017-10-27
provision/swarm: waits service update completion

andrestc:


provision/swarm/provisioner.go
```diff
@@ -939,6 +939,21 @@ func (m *serviceManager) DeployService(a provision.App, process string, labels *
 		if err != nil {
 			return errors.WithStack(err)
 		}
+		timeoutc := time.After(time.Minute)
+		for {
+			srvUpdated, err := m.client.InspectService(srvName)
+			if err != nil {
+				return errors.WithStack(err)
+			}
+			if srvUpdated.UpdateStatus.State == swarm.UpdateStateCompleted {
```
cezarsa: After an error will the update status jump directly to `UpdateStateRollback*` states without ever getting to `UpdateStateCompleted`? If so, maybe we should also `UpdateStateRollbackCompleted` as a final state.

andrestc: nice catch. updated

PR #1817, opened 2017-10-29
Handle error if app is already started

### LINKED ISSUE #1422: app-start: handle container already running
tsuru should handle the error on start a started app and raises a more readable error

```
$ tsuru app-start -a staticzito

 ---> Starting the app "staticzito"
$ tsuru app-start -a staticzito

 ---> Starting the app "staticzito"
{"Message":"","Error":"error in docker node \"http://127.0.0.1:2375\": Container already running: 0715b7defd6f284585d4391faa4e1d1f1c92d65a3c37083b64e85ac51e67a62c"}
error in docker node "http://127.0.0.1:2375": Container already running: 0715b7defd6f284585d4391faa4e1d1f1c92d65a3c37083b64e85ac51e67a62c
```

gabrielclima:
FIx #1422

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -16,6 +16,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/fsouza/go-dockerclient"
 	"github.com/pkg/errors"
 	"github.com/tsuru/tsuru/action"
 	"github.com/tsuru/tsuru/app/bind"
@@ -1866,6 +1867,9 @@ func (app *App) Start(w io.Writer, process string) error {
 	err = prov.Start(app, process)
 	if err != nil {
 		log.Errorf("[start] error on start the app %s - %s", app.Name, err)
+		if _, ok := err.(*docker.ContainerAlreadyRunning); !ok {
+			return fmt.Errorf("this application is already started and running")
+		}
 		return err
 	}
 	rebuild.RoutesRebuildOrEnqueue(app.Name)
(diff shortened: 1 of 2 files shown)


provision/docker/provisioner.go
```diff
@@ -312,6 +312,9 @@ func (p *dockerProvisioner) Start(app provision.App, process string) error {
 		return errors.New(fmt.Sprintf("Got error while getting app containers: %s", err))
 	}
 	err = runInContainers(containers, func(c *container.Container, _ chan *container.Container) error {
+		if c.Status == provision.StatusStarted.String() {
```
guilhermebr: Can you add a test to this case?

app/app.go
```diff
@@ -1866,6 +1867,10 @@ func (app *App) Start(w io.Writer, process string) error {
 	err = prov.Start(app, process)
 	if err != nil {
 		log.Errorf("[start] error on start the app %s - %s", app.Name, err)
+		if _, ok := err.(*docker.ContainerAlreadyRunning); !ok {
```
cezarsa: We shouldn't have docker specific errors in the `app` package as especially since this is only implemented in the `provision/docker` provisioner and would not work as expected when using kubernetes or swarm provisioners.
If we're to handle already running errors in the `app` package a new error could be created in the `provision` package, something like `provision.ErrAppAlreadyStarted` and all provisioners would be able to return it.

provision/docker/provisioner.go
cezarsa: Also, I don't think we should blindly trust on the information provided by MongoDB as it might be stale, the start operation should still be issued to the containers and we could check the returned `startErr` and handle it accordingly.

guilhermebr:
@gabrielclima are you still planning to work this?

gabrielclima:
@guilhermebr yes, i'll finish this PR. 
Sorry about the delay.

PR #1851, opened 2017-11-27
integration: uses iptables to remove conectivity between node and api

andrestc:
Using iptables instead of machine-destroy to force the healing of
a allows using kubernetes with a cloudprovider that keeps track of nodes
and machines, since removing the machine with `machine ddestroy` would
cause the node to be removed from the kubernetes cluster by the
cloudprovider (this causes the machine to never get healed as tsuru has
no way to known that such node existed in the first place).

### DIFF
--- a/integration/install_test.go
+++ b/integration/install_test.go
@@ -102,13 +102,17 @@ func installerComposeTest() ExecFlow {
 
 func installerTest() ExecFlow {
 	flow := ExecFlow{
-		provides: []string{"targetaddr"},
+		provides: []string{"targetaddr", "installerhostname"},
 	}
 	flow.forward = func(c *check.C, env *Environment) {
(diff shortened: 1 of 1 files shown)


integration/install_test.go
```diff
@@ -244,14 +248,20 @@ func nodeHealer() ExecFlow {
 			}
 		}
 		c.Assert(machineID, check.Not(check.Equals), "")
-		res = T("machine-destroy", machineID).Run(env)
+		regex := regexp.MustCompile(`(?:https?:\/\/)?(\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})(?::\d+)?`)
+		parts := regex.FindStringSubmatch(nodeAddr)
+		c.Assert(parts, check.HasLen, 2)
+		nodeIP := parts[1]
```
cezarsa: I think you could use `net.URLToHost(nodeAddr)` here instead of a regex.

PR #1852, opened 2017-11-27
provision/kubernetes: add headless svc to apps and subdomain to dep pods

cezarsa:


provision/kubernetes/helpers.go
```diff
@@ -45,6 +45,12 @@ func deploymentNameForApp(a provision.App, process string) string {
 	return fmt.Sprintf("%s-%s", name, process)
 }
 
+func headlessServiceNameForApp(a provision.App, process string) string {
```
ggarnier: You could add a test for this function

cezarsa: Done!

PR #1857, opened 2017-11-29
iaas/dockermachine: ensure machine name is valid hostname

fsouza:
Since the iaas implementation derives the machine name from the pool
and tsuru doesn't ensure that the pool is a valid hostname, the
implementation must strip any invalid character from the pool name
before generating the machine name.


cezarsa:
Thanks! :)

PR #1863, opened 2017-11-30
action: handle panics as errors during pipelines

andrestc:
Ideally, one action should never panic but this change makes sure we
handle it gracefully (calling the backwards function as when an error occurs)
to minimize the chance of inconsistences.

### DIFF
--- a/action/action.go
+++ b/action/action.go
@@ -5,6 +5,7 @@
 package action
 
 import (
+	"fmt"
 	"sync"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


action/action.go
```diff
@@ -132,16 +133,27 @@ func (p *Pipeline) Result() Result {
...
 	if len(p.actions) == 0 {
 		return ErrPipelineNoActions
 	}
 	fwCtx := FWContext{Params: params}
-	for i, a := range p.actions {
+	var i int
+	var a *Action
+	defer func() {
+		if r := recover(); r != nil {
+			log.Debugf("[pipeline] PANIC running the Forward for the %s action - %v", a.Name, r)
```
cezarsa: I think `log.Errorf` would be more appropriate here.

PR #1864, opened 2017-11-30
api: add node-info endpoint

guilhermebr:


api/node.go
```diff
@@ -675,3 +675,68 @@ func rebalanceNodesHandler(w http.ResponseWriter, r *http.Request, t auth.Token)
...
+// responses:
+//   200: Ok
+//   400: Invalid data
+//   404: Not found
+func infoNodeHandler(w http.ResponseWriter, r *http.Request, t auth.Token) error {
+	address := r.URL.Query().Get(":address")
+	if address == "" {
+		return errors.Errorf("Node address is required.")
+	}
+	if err := validateNodeAddress(address); err != nil {
```
cezarsa: I don't think this validate call is necessary. Actually it can cause errors if the address is not a URL as is the case for kubernetes provisioner nodes.

api/node.go
```diff
@@ -675,3 +675,68 @@ func rebalanceNodesHandler(w http.ResponseWriter, r *http.Request, t auth.Token)
 	fmt.Fprintf(writer, "Units successfully rebalanced!\n")
 	return nil
 }
+
+type infoNodeResponse struct {
```
cezarsa: If this struct were public it could be used do deserialize data in the client.

PR #1876, opened 2017-12-07
provision/pool: validates constraint type on append

andrestc:
This PR also includes a refactor of constraints fields (now called types).

### DIFF
--- a/provision/pool/pool.go
+++ b/provision/pool/pool.go
@@ -69,7 +69,7 @@ func (p *Pool) GetTeams() ([]string, error) {
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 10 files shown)


provision/pool/pool.go
```diff
@@ -189,7 +186,7 @@ func (p *Pool) allowedValues() (map[string][]string, error) {
 				validNames = append(validNames, n)
 			}
 		}
-		resolved[k] = validNames
+		resolved[string(k)] = validNames
```
ggarnier: Shouldn't `allowedValues` return values change to `map[poolConstraintType][]string, error`?

PR #1881, opened 2017-12-11
Image deploy hooks

ggarnier:
Set deploy hooks when deploying from an image

### DIFF
--- a/builder/docker/builder.go
+++ b/builder/docker/builder.go
@@ -143,6 +143,13 @@ func loadTsuruYaml(client provision.BuilderDockerClient, app provision.App, imag
 	}
 	customData := map[string]interface{}{
 		"healthcheck": tsuruYamlData.Healthcheck,
+		"hooks": map[string]interface{}{
+			"build": tsuruYamlData.Hooks.Build,
+			"restart": map[string]interface{}{
+				"before": tsuruYamlData.Hooks.Restart.Before,
+				"after":  tsuruYamlData.Hooks.Restart.After,
+			},
+		},
 	}
 	return customData, err
 }
--- a/builder/docker/builder.go
+++ b/builder/docker/builder.go
@@ -170,6 +170,8 @@ func runBuildHooks(client provision.BuilderDockerClient, app provision.App, imag
 	}
 
 	cmd := strings.Join(buildHooks, " && ")
+	fmt.Fprintln(evt, "---- Running build hooks ----")
+	fmt.Fprintf(evt, " ---> Running %q\n", cmd)
 	containerID, err := runCommandInContainer(client, evt, imageID, cmd, app, evt, nil)
 	if err != nil {
 		return "", err
--- a/builder/docker/builder.go
+++ b/builder/docker/builder.go
@@ -103,7 +103,7 @@ func imageBuild(client provision.BuilderDockerClient, app provision.App, imageID
 	fmt.Fprintln(evt, "---- Getting process from image ----")
 	cmd := "(cat /home/application/current/Procfile || cat /app/user/Procfile || cat /Procfile || true) 2>/dev/null"
 	var procfileBuf bytes.Buffer
-	err := runCommandInContainer(client, evt, imageID, cmd, app, &procfileBuf, nil)
+	_, err := runCommandInContainer(client, evt, imageID, cmd, app, &procfileBuf, nil)
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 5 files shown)


builder/docker/builder.go
```diff
@@ -172,16 +223,50 @@ func runCommandInContainer(client provision.BuilderDockerClient, evt *event.Even
 	}
 	waiter, err := client.AttachToContainerNonBlocking(attachOptions)
 	if err != nil {
-		return err
+		return "", err
```
cezarsa: We should return `cont.ID` both here and after a StartContainer error to allow the container to be removed even after a failure. Another possibility is removing the container inside this function if we are to return an error.

builder/docker/builder.go
```diff
@@ -120,46 +127,90 @@ func imageBuild(client provision.BuilderDockerClient, app provision.App, imageID
...
+}
+
+func tsuruYamlToCustomData(yaml *provision.TsuruYamlData) map[string]interface{} {
+	if yaml == nil {
+		return nil
+	}
+
+	return map[string]interface{}{
+		"healthcheck": yaml.Healthcheck,
+		"hooks": map[string]interface{}{
```
cezarsa: I think having `"hooks": yaml.Hooks,` would work here.

builder/docker/builder.go
```diff
@@ -120,46 +127,90 @@ func imageBuild(client provision.BuilderDockerClient, app provision.App, imageID
...
 	return newImage, nil
 }
 
-func loadTsuruYaml(client provision.BuilderDockerClient, app provision.App, imageID string, evt *event.Event) (map[string]interface{}, error) {
+func loadTsuruYaml(client provision.BuilderDockerClient, app provision.App, imageID string, evt *event.Event) (*provision.TsuruYamlData, string, error) {
 	path := defaultArchivePath + "/current"
 	cmd := fmt.Sprintf("(cat %s/tsuru.yml || cat %s/tsuru.yaml || cat %s/app.yml || cat %s/app.yaml || true) 2>/dev/null", path, path, path, path)
 	var buf bytes.Buffer
-	err := runCommandInContainer(client, evt, imageID, cmd, app, &buf, nil)
+	containerID, err := runCommandInContainer(client, evt, imageID, cmd, app, &buf, nil)
```
cezarsa: What do you think about moving the `defer RemoveContainer(containerID)` call to this function instead of returning the ID and making the parent remove it?

builder/docker/builder.go
```diff
@@ -120,46 +127,90 @@ func imageBuild(client provision.BuilderDockerClient, app provision.App, imageID
...
+	fmt.Fprintf(evt, " ---> Running %q\n", cmd)
+	containerID, err := runCommandInContainer(client, evt, imageID, cmd, app, evt, nil)
+	if err != nil {
+		return "", err
+	}
+
+	repo, tag := splitImageName(imageID)
+	opts := docker.CommitContainerOptions{
+		Container:  containerID,
+		Repository: repo,
```
cezarsa: I think the only reason why `Repository` and `Tag` is needed here is because of a bug in docker-cluster, as it isn't storing information about the image if they are not present. Let's leave it like this for now but we may be able to simplify this in the future.

builder/docker/builder.go
ggarnier: The problem is that I call `runCommandInContainer` also in `runBuildHooks` function, and I can't remove it there, because it's commited

cezarsa: Oh right, you can only remove it after it's been pushed. 👍

ggarnier:
@cezarsa PTAL

builder/docker/builder.go
```diff
@@ -120,46 +128,81 @@ func imageBuild(client provision.BuilderDockerClient, app provision.App, imageID
...
+	if tsuruYamlData == nil || len(tsuruYamlData.Hooks.Build) == 0 {
+		return "", nil
+	}
+
+	cmd := strings.Join(tsuruYamlData.Hooks.Build, " && ")
+	fmt.Fprintln(evt, "---- Running build hooks ----")
+	fmt.Fprintf(evt, " ---> Running %q\n", cmd)
+	containerID, err := runCommandInContainer(client, evt, imageID, cmd, app, evt, evt)
+	if err != nil {
+		return "", err
```
cezarsa: Return containerID here to ensure removal in case of partial failure.

PR #1887, opened 2017-12-18
Enabling router opts on API routers

gfleury:
Enable routers of type API to have routerOpts interface. 
This will allow us to use specific parameters on API routers, as using kubernetes-routes to get some opts and use as 'Host' on ingress to be consumed by nginx-ingress.


cezarsa:
Nice catch, thank you!

PR #1889, opened 2017-12-20
Add support to run 2 server instances (HTTP/HTTPS)

ggarnier:
With this PR, it's possible to start 2 tsuru server instances simultaneously, one with HTTP and the other with HTTPS. The use case for this is when you want to migrate from HTTP to HTTPS, and you need to keep them both running for some time.

To enable HTTP server in port 8080 and HTTPS server in port 8443, for instance, set this in `tsuru.conf`:

```
listen: :8080
use-tls: true
tls:
  listen: :8443
  cert-file: test.crt
  key-file: test.key
```

To use only HTTPS, set `use-tls: true` and only one of `listen`, `tls:listen` keys.


api/server_test.go
```diff
@@ -77,3 +123,86 @@ func (s *S) TestIsNotAdmin(c *check.C) {
...
+func (s *S) TestRunHTTPServer(c *check.C) {
+	port, err := selectAvailablePort()
+	c.Assert(err, check.IsNil)
+	config.Set("listen", "0.0.0.0:"+port)
+	config.Set("queue:mongo-url", "127.0.0.1:27017")
+	config.Set("queue:mongo-database", "queuedb")
+	RegisterHandler("/foo", "GET", AuthorizationRequiredHandler(authorizedTsuruHandler))
+	defer resetHandlers()
+
+	go RunServer(false)
```
cezarsa: Calling RunServer with `false` for the `dryMode` is quite problematic, this is the cause of all the race conditions in the tests and this happens because the test isn't closing the old server and calling shutdown on multiple initialized components. Calling `shutdown.Do()` before the test exits should be able to fix this. However, I'm not sure if some of our packages can safely handle being reinitialized after a shutdown.

api/server.go
```diff
@@ -560,19 +569,51 @@ func startServer(handler http.Handler) {
...
+			Addr:         tlsListen,
+			Handler:      handler,
 		}
+		shutdown.Register(httpsSrv)
 	}
-	<-shutdownChan
+
+	if httpsSrv != nil {
+		if httpSrv != nil {
+			go startHttpServer(httpSrv, listen)
```
cezarsa: By starting a goroutine here we are ignoring possible errors during listening. Maybe we should have a channel for errors to capture then, something like:

```
errCh := make(chan error, 2)
if httpsSrv != nil {
  go func() { errCh <- startHttpsServer(httpsSrv, listen, ...) }()
}
if httpSrv != nil {
  go func() { errCh <- startHttpServer(httpSrv, listen) }()
}
return <-errCh
```

Also, I don't think we should call `fatal` in `createServers` since we're able to return errors. We should simply return the errors and let the caller call `fatal` for any returned error.

PR #1892, opened 2017-12-25
Openapi router api specification, 

gfleury:
I've looked for something similar in the repo but didn't found anything. Please point me if there is any api spec documented, besides generating code with impl. 
Easy to implement new routers in other languanges generating with swagger.
We should also have one for 'repo-manager' api and 'service' api. 
Right now its hard to start integrating because you need to look into examples/original code.


cezarsa:
Thank you for working on this.

PR #1897, opened 2017-12-27
docs: adds openapi reference for service api

andrestc:
Continues the work started by #1892

### DIFF
--- a/docs/reference/service_api.yaml
+++ b/docs/reference/service_api.yaml
@@ -1,0 +1,302 @@
+#
+# Tsuru Service API Specification
+#
+openapi: 3.0.0
+
+info:
+  title: Tsuru Service API
+  description: Tsuru Service API used to integration with external services
+  version: "1.0.0"
+
+paths:
+  /resources/plans:
+    get:
+      summary: List plans available
+      description: |
+        The service endpoint returns available service plans
+      responses:
+        200:
+          description: An array of plans
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/Plan'
+        default:
+          $ref: '#/components/schemas/Error'
+  /resorces:
+    post:
+      summary: Create a new service instance
+      description: |
+        The service endpoint creates a new service instance
+      requestBody:
+        required: true
+        content:
+          application/x-www-form-urlencoded:
+            schema:
+              $ref: '#/components/schemas/InstanceForm'
+      responses:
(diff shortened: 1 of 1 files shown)


docs/reference/service_api.yaml
```diff
@@ -0,0 +1,302 @@
...
+        required: true
+        schema:
+          type: string
+    get:
+      summary: Service instance additional information
+      description: |
+        The service endpoint returns additional information about the instance
+      responses:
+        404:
+          description: No additional information available
```
cezarsa: Doesn't 404 here mean instance not found?

docs/reference/service_api.yaml
```diff
@@ -0,0 +1,302 @@
...
+        app-host:
+          type: string
+          description: App host
+        app-name:
+          type: string
+          description: App name
+      required:
+        - app-host
+        - app-name
+    InstanceInfo:
```
cezarsa: Taking a quick look at the code it looks like the InstanceInfo object is actually a map[string]string, do we really only accept `label` and `value` keys?

docs/reference/service_api.yaml
andrestc: I don't think so. Ref: https://docs.tsuru.io/stable/services/api.html#additional-info-about-an-instance
`404: when the API doesn’t have extra info about the service instance. There’s no need to include anything in the response body.`
Also: https://github.com/tsuru/tsuru/blob/master/service/endpoint.go#L371

docs/reference/service_api.yaml
andrestc: Yeah, I got misled by the docs: https://docs.tsuru.io/stable/services/api.html#additional-info-about-an-instance
I'll fix this.

docs/reference/service_api.yaml
cezarsa: Ugh okay, if this is what the docs say let's keep it like this. But without this assurance we have no way to check if a given instance exists in the service. :(

docs/reference/service_api.yaml
```diff
@@ -0,0 +1,313 @@
...
+          description: An array of plans
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/Plan'
+        default:
+          $ref: '#/components/schemas/Error'
+  /resorces:
```
cezarsa: Typo: `resoUrces`

PR #1902, opened 2017-12-28
Mock builder

guilhermebr:
Create MockBuilder and Replace FakeBuilder


api/deploy_test.go
```diff
@@ -91,7 +90,7 @@ func (s *DeploySuite) SetUpSuite(c *check.C) {
 	c.Assert(err, check.IsNil)
 	s.logConn, err = db.LogConn()
 	c.Assert(err, check.IsNil)
-	s.builder = fake.NewFakeBuilder()
+	s.builder = &builder.MockBuilder{}
```
cezarsa: Wouldn't it be safer to move this initialization to `SetUpTest`? Declaring it here makes it possible for a callback set by one test to leak to another test. The same applies to all other suites with mock instances.

guilhermebr: Sure, I can move this to SetUpTest and remove the Reset() method

PR #1933, opened 2018-02-05
provision/kubernetes: use attach instead of relying on kube logs

andrestc:
This commit changes the runPod func to use attach instead of relying
on kubernetes logs to fetch the output of the command that ran
in the given pod/containers. This allows the image deploy to work
on kubernetes running nodes without the default json file loging.


provision/kubernetes/helpers.go
```diff
@@ -558,10 +559,12 @@ func runPod(args runSinglePodArgs) error {
 			RestartPolicy: apiv1.RestartPolicyNever,
 			Containers: []apiv1.Container{
 				{
-					Name:    args.name,
-					Image:   args.image,
-					Command: args.cmds,
-					Env:     args.envs,
+					Name:      args.name,
+					Image:     args.image,
+					Command:   []string{"sh", "-c", "cat >/dev/null && " + args.cmd},
```
cezarsa: I think we can use `-ec` here, `imageTagAndPush` used to rely on `-e` to stop after a possible docker pull error.

PR #1943, opened 2018-02-22
provision/docker: fix scheduler filtering when scheduleAnyNode is used

cezarsa:


provision/docker/scheduler_test.go
```diff
@@ -138,6 +138,78 @@ func (s *S) TestSchedulerScheduleFilteringNodes(c *check.C) {
...
 }
 
+func (s *S) TestFilterNodes(c *check.C) {
+	tests := []struct {
+		nodes    []cluster.Node
+		filter   map[string]struct{}
+		expected []cluster.Node
+	}{
+		{
+			nodes: []cluster.Node{
```
ggarnier: This test case is duplicated

cezarsa: Fixed

PR #1949, opened 2018-03-02
Team service refactoring

ggarnier:
In this PR, I split `TeamService` in 2 layers: `TeamService` contains the domain business rules, and `TeamStorage` is the persistance layer. With this, we could use `TeamService` with another implementation of storage.

I also added a mock implementation of `TeamService`, so api tests can use it instead of testing with all the layers, including the DB.

The `servicemanager` pkg with global variables should be a temporary solution, while we don't migrate all the code to this model.


api/server.go
```diff
@@ -52,12 +53,17 @@ type TsuruHandler struct {
...
+	Team authTypes.TeamService
+}
+
 func fatal(err error) {
 	fmt.Println(err.Error())
 	log.Fatal(err.Error())
 }
 
 var tsuruHandlerList []TsuruHandler
+var ServiceManager *serviceManager
```
guilhermebr: Why this global var instead of using servicemanager package like in others packages?

ggarnier: Thanks, I forgot to remove it.

ggarnier:
Refs #1994

PR #1964, opened 2018-03-14
Team service adjustments

ggarnier:
- moves `TeamStorage` type to `types/auth`
- creates `MockTeamStorage` for TeamService tests


ggarnier:
Refs #1994

PR #1966, opened 2018-03-15
app/plan,integration: DefaultPlan creates autogenerated plan when empty

guilhermebr:


app/plan.go
```diff
@@ -51,7 +53,32 @@ func (s *planService) FindByName(name string) (*appTypes.Plan, error) {
 
 // DefaultPlan implements DefaultPlan method of PlanService interface
 func (s *planService) DefaultPlan() (*appTypes.Plan, error) {
-	return s.storage.FindDefault()
+	plan, err := s.storage.FindDefault()
+	if err != nil && err == appTypes.ErrPlanDefaultNotFound {
+		plans, err := s.storage.FindAll()
+		if err != nil {
+			log.Errorf("Error listing Plans: %s", err)
+			return nil, appTypes.ErrPlanDefaultNotFound
```
cezarsa: I think it would be better to return the original error here, returning plan not found may mislead the end user.

app/plan.go
```diff
@@ -51,7 +53,32 @@ func (s *planService) FindByName(name string) (*appTypes.Plan, error) {
...
+				Name:     "autogenerated",
+				Memory:   int64(configMemory) * 1024 * 1024,
+				Swap:     int64(configSwap-configMemory) * 1024 * 1024,
+				CpuShare: 100,
+				Default:  true,
+			}
+			err := s.storage.Insert(df)
+			if err != nil {
+				log.Errorf("Error inserting autogenerated Default Plan: %s", err)
+				return nil, appTypes.ErrPlanDefaultNotFound
```
cezarsa: I also think the original error should be returned here

app/plan.go
```diff
@@ -51,7 +53,32 @@ func (s *planService) FindByName(name string) (*appTypes.Plan, error) {
...
-	return s.storage.FindDefault()
+	plan, err := s.storage.FindDefault()
+	if err != nil && err == appTypes.ErrPlanDefaultNotFound {
+		plans, err := s.storage.FindAll()
+		if err != nil {
+			log.Errorf("Error listing Plans: %s", err)
+			return nil, appTypes.ErrPlanDefaultNotFound
+		}
+		if len(plans) == 0 {
+			configMemory, _ := config.GetInt("docker:memory")
```
ggarnier: Just a detail, I would extract the default plan creation to a separate private function, to let the code easier do read.

app/plan.go
```diff
@@ -51,7 +53,32 @@ func (s *planService) FindByName(name string) (*appTypes.Plan, error) {
 
 // DefaultPlan implements DefaultPlan method of PlanService interface
 func (s *planService) DefaultPlan() (*appTypes.Plan, error) {
-	return s.storage.FindDefault()
+	plan, err := s.storage.FindDefault()
```
cezarsa: To reduce the indentation levels and avoid shadowing the err variable inside the if below I think we could try exiting earlier, as soon as we find some conditions match leaving the logic always on the first level. I was thinking in something like:

```
	plan, err := s.storage.FindDefault()
	if err == nil || err != appTypes.ErrPlanDefaultNotFound {
		return plan, err
	}
	plans, err := s.storage.FindAll()
	if err != nil {
		return nil, err
	}
	if len(plans) != 0 {
		return nil, appTypes.ErrPlanDefaultNotFound
	}
	... create plan ... 
```

app/plan.go
guilhermebr: But it's returning the original error ("default plan not found")

app/plan.go
guilhermebr: But return an error about insert plan when the request is about create app don't will confuse the end user?

app/plan.go
guilhermebr: @cezarsa PTAL

Issue #1945, opened 2018-02-27
Tsuru app build latest tag after deploy


mateuspontes:
Hey guys,

When I deploy my app, Tsuru [builds](https://docs.tsuru.io/master/using/docker-image.html?highlight=build) it with commit tag, I'd like to have also `latest` tag automatically build and sent to registry. Is it possibly? With this we can use the image to run in a docker-compose like an alias.

PR #1973, opened 2018-03-22
builder,provision: adding the latest tag to deploy image

jonnyguio:
This PR is related to #1945. Now, whenever you deploy a new version of your app, you can pull the last image using the tag `latest`.

### DIFF
--- a/provision/docker/container/container.go
+++ b/provision/docker/container/container.go
@@ -18,6 +18,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/tsuru/config"
 	"github.com/tsuru/tsuru/action"
+	"github.com/tsuru/tsuru/app/image"
 	tsuruErrors "github.com/tsuru/tsuru/errors"
 	"github.com/tsuru/tsuru/event"
 	"github.com/tsuru/tsuru/log"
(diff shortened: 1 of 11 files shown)


provision/docker/container/container.go
```diff
@@ -387,48 +387,56 @@ func (c *Container) Exec(client provision.BuilderDockerClient, stdout, stderr io
...
-		if len(imgHistory) > 1 && strings.Contains(imgHistory[1].CreatedBy, "tail -f /dev/null") {
-			fullSize += imgHistory[1].Size
+	tags := []string{tag}
+	if isDeploy && tag != "latest" {
+		tags = append(tags, "latest")
+	}
+	for i, tag := range tags {
+		opts := docker.CommitContainerOptions{Container: c.ID, Repository: repository, Tag: tag}
+		done := limiter.Start(c.HostAddr)
+		image, err := client.CommitContainer(opts)
```
cezarsa: Calling `client.CommitContainer` for each tag is not necessary and it may cause a performance hit, it should be called only once (for the original tag), after the Commit is done we should call `client.TagImage` to tag the latest tag. This will also allow us to move the commit logic and the image size logic to outside the for loop (removing the ugly `if i == 0` block).
Only the `PushImage` logic should be inside this for loop.

provision/swarm/provisioner_test.go
```diff
@@ -1095,6 +1095,12 @@ func (s *S) TestDeploy(c *check.C) {
 	err := app.CreateApp(a, s.user)
 	c.Assert(err, check.IsNil)
 	attached := s.attachRegister(c, s.clusterSrv, true, a)
+	tags := []string{}
+	s.clusterSrv.CustomHandler("/images/.*/push", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		c.Assert(r.URL.Path, check.Equals, "/images/"+fmt.Sprintf("registry.tsuru.io/tsuru/app-myapp")+"/push")
```
cezarsa: No need for Sprintf() and concat here, just make it a single string.

jonnyguio: ops, fixed that

PR #1974, opened 2018-03-23
Adding Message field to DeployData/DeployInfo.

gfleury:
otherwise this field i'snt really useful as a deployment info besides being at event information


cezarsa:
We plan on eventually deprecating GET /deploys and using only the event API calls (after tsuru 2.0 whenever it may be) but for now you're right, it's useful to have it in the DeployData.

gfleury:
Hmm, ok. Makes sense then.

PR #1979, opened 2018-03-26
Remove commit override if authenticated token is from a normal user

gfleury:
This pull request is more a question than a PR. Is there any reason to override the commit id if the user is a normal user? My point is that is just meta data, what are the impacts? There no --commit option on tsuru-client, so if someone try to use it he is probably aware of what he is doing.

### DIFF
--- a/app/deploy.go
+++ b/app/deploy.go
@@ -195,7 +195,7 @@ func (o *DeployOptions) GetKind() (kind DeployKind) {
 		}
 		return DeployUpload
 	}
-	if o.Commit != "" {
+	if o.Commit != "" && o.ArchiveURL == "" {
 		return DeployGit
 	}
 	return DeployArchiveURL
--- a/api/deploy.go
+++ b/api/deploy.go
@@ -59,7 +59,6 @@ func deploy(w http.ResponseWriter, r *http.Request, t auth.Token) (err error) {
 		}
 		userName = r.FormValue("user")
 	} else {
-		commit = ""
 		userName = t.GetUserName()
 	}
 	instance, err := app.GetByName(appName)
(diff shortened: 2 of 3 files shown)


cezarsa:
I think I'm okay with allowing users to set the commit value when making the deploy API call. Tests must be fixed to match the new behavior though. @andrestc @guilhermebr @ggarnier: Any opinions on this?

gfleury:
I guess, that the difference now from origin and kind is a bit confusing for me. So, I've changed to kind being how the file is coming from and from origin being from where. But I'm not sure to be honest. From what I understood is, deployment kind and data origin. Is that right?

app/deploy.go
```diff
@@ -195,7 +195,7 @@ func (o *DeployOptions) GetKind() (kind DeployKind) {
 		}
 		return DeployUpload
 	}
-	if o.Commit != "" {
+	if o.Commit != "" && o.ArchiveURL == "" {
```
guilhermebr: Commit deploy can be made by ArchiveURL and ArchiveFile methods.
The difference is the method git pre-receive hook uses.

guilhermebr: @gfleury you can check for o.Commit and for app Token. If have both so it's of Kind DeployGit.

gfleury: Hi @guilhermebr, correct if I'm wrong, but this was the "behavior" before since o.Commit was being set to empty if the token wasn't from an app. And then rely only if commit id was empty or not to define if it was from kind DeployGit. 
Checking the token inside getKind() seems a different from how it was being done before, as getKind only rely on the DeploymentOptions to define the kind. (DeploymentOptions were being defined to force a Kind). I'm not sure how is the best way. The way I've done is following the same 'idea' from before.

Issue #1992, opened 2018-04-03
Swagger spec for tsuru API endpoints


andrestc:
We should provide a swagger 2.0 spec for the tsuru API. This spec can be used to generate clients for different languages.

A preliminary spec is in: https://github.com/tsuru/tsuru/blob/master/docs/reference/api.yaml

I'll add a list with the pending endpoints in the future, so we can track the progress.

Endpoints missing from spec:

- [ ] GET: /
- [ ] GET: /app/{}/routers
- [ ] POST: /app/{}/routers
- [ ] DELETE: /app/{}/routers/{}
- [ ] PUT: /app/{}/routers/{}
- [x] DELETE: /apps/{}
- [ ] GET: /apps/{}
- [ ] PUT: /apps/{}
- [ ] POST: /apps/{}/build
- [ ] DELETE: /apps/{}/certificate
- [ ] GET: /apps/{}/certificate
- [ ] PUT: /apps/{}/certificate
- [ ] DELETE: /apps/{}/cname
- [ ] POST: /apps/{}/cname
- [ ] POST: /apps/{}/deploy
- [ ] POST: /apps/{}/deploy/rebuild
- [ ] POST: /apps/{}/deploy/rollback
- [ ] PUT: /apps/{}/deploy/rollback/update
- [ ] POST: /apps/{}/diff
- [ ] DELETE: /apps/{}/env
- [ ] GET: /apps/{}/env
- [ ] DELETE: /apps/{}/lock
- [ ] GET: /apps/{}/log
- [ ] POST: /apps/{}/log
- [ ] GET: /apps/{}/metric/envs
- [x] GET: /apps/{}/quota
- [x] PUT: /apps/{}/quota
- [x] POST: /apps/{}/restart
- [ ] POST: /apps/{}/routes
- [ ] POST: /apps/{}/run
- [ ] GET: /apps/{}/shell
- [ ] POST: /apps/{}/sleep
- [ ] POST: /apps/{}/start
- [ ] POST: /apps/{}/stop
- [ ] DELETE: /apps/{}/teams/{}
- [ ] PUT: /apps/{}/teams/{}
- [ ] DELETE: /apps/{}/units
- [ ] PUT: /apps/{}/units
- [ ] POST: /apps/{}/units/register
- [ ] POST: /apps/{}/units/{}
- [ ] POST: /auth/login
- [ ] GET: /auth/saml
- [ ] POST: /auth/saml
- [ ] GET: /auth/scheme
[...]

PR #2000, opened 2018-04-06
docs: add user spec to swagger

jonnyguio:
Refs #1992

### DIFF
--- a/docs/reference/api.yaml
+++ b/docs/reference/api.yaml
@@ -344,9 +344,9 @@ paths:
   /1.0/users/info:
     get:
       operationId: UserInfo
-      description: Get information on an user.
+      description: Get information on logged user.
       produces:
-        - application/x-json-stream
+        - application/json
       responses:
         '200': 
           description: OK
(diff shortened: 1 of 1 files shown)


docs/reference/api.yaml
```diff
@@ -274,6 +275,408 @@ paths:
...
+          description: Not found
+          schema:
+            $ref: '#/definitions/ErrorMessage'
+      tags: 
+        - users
+      security:
+        - Bearer: []
+  /1.0/users/api-key:
+    get:
+      operationId: APITokenShow
```
andrestc: `APITokenGet` looks more common for an API.

docs/reference/api.yaml
```diff
@@ -274,6 +275,408 @@ paths:
...
+      operationId: UserSSHKeyAdd
+      description: Add SSH key to logged user.
+      parameters:
+        - name: UserSSHKeyAddData
+          in: body
+          required: true
+          schema:
+            $ref: '#/definitions/UserSSHKeyAddData'
+      consumes:
+        - x-www-form-url-encoded
```
andrestc: `application/json`

docs/reference/api.yaml
```diff
@@ -274,6 +275,408 @@ paths:
...
+          description: Not found
+          schema:
+            $ref: '#/definitions/ErrorMessage'
+      tags:
+        - users
+      security:
+        - Bearer: []
+  /1.0/users/info:
+    get:
+      operationId: UserInfo
```
andrestc: I think I'd use `UserGet` (ignoring the `/info` on the endpoint)

jonnyguio:
In my opinion, it's important to be as clear as possible what the context of the operation. Quota, for instance, is also a component of the app, so "QuotaChange" wouldn't be clear if it's about changing an user quota or an app quota. 

On the other hand, as there are no other usage of ssh keys, I think that the User part could be left out.

docs/reference/api.yaml
```diff
@@ -274,6 +274,412 @@ paths:
...
+      parameters:
+        - name: email
+          in: query
+          type: string
+          required: true
+      produces: 
+        - application/json
+      responses:
+        '200': 
+          description: <API_TOKEN>
```
andrestc: Maybe just: `API Token.`

docs/reference/api.yaml
```diff
@@ -274,6 +274,412 @@ paths:
...
+      parameters:
+        - name: email
+          in: query
+          type: string
+          required: true
+      produces:
+        - application/json
+      responses:
+        '200': 
+          description: <API_TOKEN>
```
andrestc: Same as above

docs/reference/api.yaml
```diff
@@ -274,6 +274,412 @@ paths:
...
+            $ref: '#/definitions/ErrorMessage'
+      tags:
+        - users
+      security:
+        - Bearer: []
+    post:
+      operationId: SSHKeyAdd
+      description: Add SSH key to logged user.
+      parameters:
+        - name: UserSSHKeyAddData
```
andrestc: Just `SSHKeyAddData`?

docs/reference/api.yaml
```diff
@@ -274,6 +274,412 @@ paths:
...
+    get:
+      operationId: UserGet
+      description: Get information on logged user.
+      produces:
+        - application/json
+      responses:
+        '200': 
+          description: OK
+          schema:
+            $ref: '#/definitions/UserInfoResponse'
```
andrestc: UserGetResponse

docs/reference/api.yaml
```diff
@@ -274,6 +274,412 @@ paths:
...
+          description: User e-mail.
+      consumes:
+        - application/x-www-form-urlencoded
+      produces:
+        - application/json
+      responses:
+        '200': 
+          description: OK
+          schema:
+            $ref: '#/definitions/UserQuotaViewResponse'
```
andrestc: `UserQuotaGetResponse`

docs/reference/api.yaml
```diff
@@ -274,6 +274,412 @@ paths:
...
+      description: Get quota of an user.
+      parameters:
+        - name: email
+          in: path
+          required: true
+          type: string
+          minLength: 1
+          description: User e-mail.
+      consumes:
+        - application/x-www-form-urlencoded
```
andrestc: I think we can drop the `consumes` here.

docs/reference/api.yaml
```diff
@@ -274,6 +274,412 @@ paths:
...
+      operationId: ChangePassword
+      description: Change password of logged user.
+      parameters:
+        - name: ChangePasswordData
+          in: body
+          required: true
+          schema:
+            $ref: '#/definitions/ChangePasswordData'
+      consumes:
+       - application/x-www-form-urlencoded
```
andrestc: `application/json`

PR #2005, opened 2018-04-10
docs: adds pool CRUD to spec

andrestc:
Related to #1992.

### DIFF
--- a/docs/reference/api.yaml
+++ b/docs/reference/api.yaml
@@ -439,6 +439,113 @@ paths:
         - volume
       security:
         - Bearer: []
+  /1.0/pools:
+    get:
+      operationId: PoolList
+      description: List pools.
+      produces:
+        - application/json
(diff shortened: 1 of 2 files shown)


docs/reference/api.yaml
```diff
@@ -439,6 +439,113 @@ paths:
...
+    delete:
+      operationId: PoolDelete
+      description: Deletes a pool.
+      responses:
+        '200':
+          description: Pool deleted
+        '401':
+          description: Unauthorized
+          schema:
+            $ref: '#/definitions/ErrorMessage'
```
jonnyguio: Checking the API code, the `DELETE /1.0/pool{pool}` asserts if there are no more apps in the pool to remove. If it has any, it returns FORBIDDEN (403).

andrestc: done

PR #2018, opened 2018-04-20
Add support for webhooks triggered on events

cezarsa:
WebHooks can now be registered in the tsuru API to be triggered whenever an event matches the webhook specification.

The two main parts for this PR are:
1 - Routes for WebHooks CRUD and corresponding service and storage methods.
2 - Each time an event is finished (`evt.Done()` call), tsuru will send to event to a channel where a dedicated goroutine will pick the event and search for all webhooks matching it. After listing all webhooks that must be called tsuru will proceed to call them sequentially.

By default the webhook request is a POST containing in its body the event that triggered it serialized as JSON, but this behavior can be overridden by setting a custom method and body in the event.

### DIFF
--- a/api/webhook.go
+++ b/api/webhook.go
@@ -40,7 +40,33 @@ func webhookList(w http.ResponseWriter, r *http.Request, t auth.Token) error {
 		w.WriteHeader(http.StatusNoContent)
 		return nil
 	}
+	w.Header().Set("Content-Type", "application/json")
 	return json.NewEncoder(w).Encode(webhooks)
+}
+
+// title: webhook info
+// path: /events/webhooks/{name}
+// method: GET
+// produce: application/json
+// responses:
+//   200: Get webhook
+//   404: Not found
+//   401: Unauthorized
+func webhookInfo(w http.ResponseWriter, r *http.Request, t auth.Token) error {
+	webhookName := r.URL.Query().Get(":name")
+	webhook, err := servicemanager.WebHook.Find(webhookName)
+	if err != nil {
+		if err == eventTypes.ErrWebHookNotFound {
+			w.WriteHeader(http.StatusNotFound)
+		}
+		return err
+	}
+	ctx := permission.Context(permission.CtxTeam, webhook.TeamOwner)
+	if !permission.Check(t, permission.PermWebhookRead, ctx) {
+		return permission.ErrUnauthorized
+	}
+	w.Header().Set("Content-Type", "application/json")
+	return json.NewEncoder(w).Encode(webhook)
 }
 
 // title: webhook create
--- a/api/webhook.go
+++ b/api/webhook.go
@@ -1,3 +1,7 @@
+// Copyright 2018 tsuru authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
 package api
 
 import (
@@ -21,7 +25,6 @@ import (
 // responses:
 //   200: List webhooks
 //   204: No content
-//   401: Unauthorized
 func webhookList(w http.ResponseWriter, r *http.Request, t auth.Token) error {
 	ctxs := permission.ContextsForPermission(t, permission.PermWebhookRead, permission.CtxTeam)
 	var teams []string
@@ -32,7 +35,7 @@ func webhookList(w http.ResponseWriter, r *http.Request, t auth.Token) error {
 		}
 		teams = append(teams, c.Value)
 	}
-	webhooks, err := servicemanager.WebHook.List(teams)
+	webhooks, err := servicemanager.Webhook.List(teams)
 	if err != nil {
 		return err
 	}
@@ -54,9 +57,9 @@ func webhookList(w http.ResponseWriter, r *http.Request, t auth.Token) error {
 //   401: Unauthorized
 func webhookInfo(w http.ResponseWriter, r *http.Request, t auth.Token) error {
 	webhookName := r.URL.Query().Get(":name")
-	webhook, err := servicemanager.WebHook.Find(webhookName)
+	webhook, err := servicemanager.Webhook.Find(webhookName)
 	if err != nil {
-		if err == eventTypes.ErrWebHookNotFound {
+		if err == eventTypes.ErrWebhookNotFound {
 			w.WriteHeader(http.StatusNotFound)
 		}
 		return err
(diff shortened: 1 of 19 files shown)


api/webhook.go
```diff
@@ -0,0 +1,202 @@
...
+		return err
+	}
+	defer func() {
+		evt.Done(err)
+	}()
+	err = servicemanager.WebHook.Update(webhook)
+	if err == eventTypes.ErrWebHookNotFound {
+		w.WriteHeader(http.StatusNotFound)
+	} else if _, ok := err.(*errors.ValidationError); ok {
+		w.WriteHeader(http.StatusBadRequest)
```
andrestc: This is already handled by the middleware: https://github.com/tsuru/tsuru/blob/master/api/middleware.go#L119

api/webhook.go
```diff
@@ -0,0 +1,202 @@
...
+		return err
+	}
+	defer func() {
+		evt.Done(err)
+	}()
+	err = servicemanager.WebHook.Create(webhook)
+	if err == eventTypes.ErrWebHookAlreadyExists {
+		w.WriteHeader(http.StatusConflict)
+	} else if _, ok := err.(*errors.ValidationError); ok {
+		w.WriteHeader(http.StatusBadRequest)
```
andrestc: This is already handled by the middleware: https://github.com/tsuru/tsuru/blob/master/api/middleware.go#L119

docs/reference/api.yaml
```diff
@@ -1695,6 +1695,138 @@ paths:
...
+  /1.6/events/webhooks/{name}:
+    parameters:
+      - name: name
+        in: path
+        required: true
+        type: string
+        minLength: 1
+        description: WebHook name.
+    get:
+      operationId: WebHookInfo
```
andrestc: I think we agreed to use `Get` here, e.g: `WebHookGet`

event/event.go
```diff
@@ -243,6 +245,26 @@ func GetTargetType(t string) (TargetType, error) {
 		return TargetTypeTeam, nil
 	case "user":
 		return TargetTypeUser, nil
+	case "iaas":
```
andrestc: nit: I think we could refactor this function to use a map string -> TargetType

event/webhook/webhook.go
```diff
@@ -0,0 +1,259 @@
...
+
+type webHookService struct {
+	storage eventTypes.WebHookStorage
+	evtCh   chan string
+	quitCh  chan struct{}
+	doneCh  chan struct{}
+}
+
+func (s *webHookService) Shutdown(ctx context.Context) error {
+	doneCtx := ctx.Done()
```
andrestc: Just curious: any specific reason for this, instead of calling <-ctx.Done() from within the select statement?

api/webhook.go
```diff
@@ -0,0 +1,202 @@
+package api
```
ggarnier: [nit] copyright comment missing

event/webhook/webhook_test.go
```diff
@@ -0,0 +1,258 @@
+// Copyright 2017 tsuru authors. All rights reserved.
```
ggarnier: [nit] 2018

storage/mongodb/webhook.go
```diff
@@ -0,0 +1,129 @@
...
+	mgo "github.com/globalsign/mgo"
+	"github.com/globalsign/mgo/bson"
+	"github.com/tsuru/tsuru/db"
+	dbStorage "github.com/tsuru/tsuru/db/storage"
+	"github.com/tsuru/tsuru/types/event"
+)
+
+type webhookStorage struct{}
+
+func webhoookCollection(conn *db.Storage) *dbStorage.Collection {
```
ggarnier: Typo in `webhoookCollection`

storage/mongodb/webhook.go
```diff
@@ -0,0 +1,129 @@
...
+	if teams != nil {
+		query = bson.M{"teamowner": bson.M{"$in": teams}}
+	}
+	return s.findQuery(query)
+}
+
+func (s *webhookStorage) FindByEvent(f event.WebHookEventFilter, isSuccess bool) ([]event.WebHook, error) {
+	for _, name := range f.KindNames {
+		parts := strings.Split(name, ".")
+		if len(parts) == 0 {
```
ggarnier: Is there a case where `len(parts)` is 0?

event/webhook/webhook.go
```diff
@@ -0,0 +1,259 @@
...
+		evtCh:   make(chan string, chanBufferSize),
+		quitCh:  make(chan struct{}),
+		doneCh:  make(chan struct{}),
+	}
+	go s.run()
+	shutdown.Register(s)
+	return s, nil
+}
+
+type webHookService struct {
```
ggarnier: You're mixing `Webhook` and `WebHook` cases. We should choose one of them

cezarsa: Going with `Webhook`

storage/mongodb/webhook.go
cezarsa: Nope, removing this

storage/mongodb/webhook.go
cezarsa: 👍

event/webhook/webhook_test.go
cezarsa: 👍

api/webhook.go
cezarsa: 👍

event/webhook/webhook.go
cezarsa: Nope, no reason, moving it

event/event.go
cezarsa: Yeah, ugly. I think this should be either generated or removed, I'm not sure if we really need this GetTargetType anymore. I'll address this in another PR.

docs/reference/api.yaml
cezarsa: 👍

api/webhook.go
cezarsa: 👍

api/webhook.go
cezarsa: 👍

api/webhook.go
```diff
@@ -0,0 +1,202 @@
...
+)
+
+// title: webhook list
+// path: /events/webhooks
+// method: GET
+// produce: application/json
+// responses:
+//   200: List webhooks
+//   204: No content
+//   401: Unauthorized
```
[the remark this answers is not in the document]
cezarsa: 👍

event/event.go
andrestc: 👍

PR #2030, opened 2018-04-27
provision/kubernetes: support deploy cancelation

andrestc:
This implements cancelation for kubernetes deploys.

### DIFF
--- a/provision/kubernetes/helpers.go
+++ b/provision/kubernetes/helpers.go
@@ -124,9 +124,9 @@ func waitFor(ctx context.Context, fn func() (bool, error), onCancel func() error
 		select {
 		case <-ctx.Done():
 			if onCancel == nil {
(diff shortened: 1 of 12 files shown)


provision/kubernetes/helpers.go
```diff
@@ -121,11 +122,11 @@ func waitFor(timeout time.Duration, fn func() (bool, error), onTimeout func() er
 			return nil
 		}
 		select {
-		case <-timeoutCh:
-			if onTimeout == nil {
-				err = errors.Errorf("timeout after %v", timeout)
+		case <-ctx.Done():
+			if onCancel == nil {
+				err = errors.Errorf("canceled after %v", time.Since(start))
```
cezarsa: I think we should call `err = errors.Wrapf(ctx.Err(), ...` here. This way we would be able to differentiate between DeadlineExceeded and explicit cancelation of events.

Issue #1994, opened 2018-04-04
Isolate storage layer


ggarnier:
We should keep all interactions with MongoDB in a separate package. With this, we would decouple from MongoDB and be able to support other database implementations. Also, we could use mocks in unit tests, making them faster and removing this dependency from unit tests in other packages.

PR #2031, opened 2018-04-30
Refactor QuotaStorage and Create QuotaService

jonnyguio:
Refs #1994 

Another work on the refactor of the app. Both User and App quota has been refactored.


app/actions.go
```diff
@@ -89,7 +91,9 @@ var insertApp = action.Action{
 			return nil, err
 		}
 		defer conn.Close()
-		app.Quota = quota.Unlimited
+		if app.Quota == (appTypes.Quota{}) {
```
ggarnier: Shouldn't you just set `app.Quota.Limit = quota.Unlimited`?

app/actions_test.go
```diff
@@ -5,18 +5,17 @@
...
 	"github.com/tsuru/tsuru/action"
 	"github.com/tsuru/tsuru/app/bind"
 	"github.com/tsuru/tsuru/auth"
-	"github.com/tsuru/tsuru/quota"
 	"github.com/tsuru/tsuru/repository"
 	"github.com/tsuru/tsuru/router/routertest"
 	appTypes "github.com/tsuru/tsuru/types/app"
-	"gopkg.in/check.v1"
+	authTypes "github.com/tsuru/tsuru/types/auth"
+	check "gopkg.in/check.v1"
```
ggarnier: You don't need to set the import name here, `check` is the default.

jonnyguio: I need double `check` with the Go automatic imports.

app/actions.go
jonnyguio: Yes. Changing to reference `UnlimitedQuota`, both in `app` and in `quota`. Now to create a generic unlimited quota just copy from that struct.

app/quota.go
```diff
@@ -5,114 +5,109 @@
...
-	app, err := checkAppLimit(app.Name, quantity)
+type appQuotaService struct {
+	storage appTypes.QuotaStorage
+}
+
+// ReserveUnits implements ReserveUnits method from AppQuotaService interface
+func (s *appQuotaService) ReserveUnits(appName string, quantity int) error {
+	quota, err := s.storage.FindByAppName(appName)
 	if err != nil {
+		if err == appTypes.ErrAppNotFound {
```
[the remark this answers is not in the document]
jonnyguio: I agree that should just keep one. Just thought it could be out of the scope of this PR. Since it's just changing some references without any major logic change, I'm moving forward with changing to only the one in `appTypes`.

cezarsa:
I'm approving this for now, but I think we should revisit this code in the future and try to simplify this code.

What bothers me most is the impact this PR has in the codebase (`+1,742 −734`) with no apparent benefit and no new features at all. This is a refactor that adds complexity and makes it harder to maintain the codebase.

By looking at the code the added complexity seems to be two-fold. First we have increased test complexity by defining mock callbacks with the expected behavior for each storage call on every test, this adds up fast and in my opinion it would be far simpler to use a real storage implementation in the tests. The second reason is more complex. I think the Quota interfaces for the 2 services (AppQuotaService, UserQuotaService) are too big and finding a new abstraction for the concept of Quotas would allow us to simplify the interface reducing the number of methods and ideally having a single Service interface that can be used for both app and user quotas.

Thinking about Quotas conceptually they can be defined as "Maximum number of items of a type X per object Y". This way user quotas would be "Maximum number of items of type App per object User" and app quotas would be "Maximum number of items of type Unit per object App". This abstraction might allow us to have a single interface that would something like:

```
type QuotaService interface {
  Set(type, object string, limit int)
  Get(type, object string) int
  Inc(type, object string)
  ...
}
```

This would be quite a departure from today's model but I think in the end it would allow us to simplify code all around and make it easier to understand.

PR #2032, opened 2018-04-30
compose: adds local Dockerfile for build

andrestc:
This PR adds a local Dockerfile to be used by compose-build when building tsuru
for development. Our alpine platforms are not really maintained right now and
updating Go was getting harder and harder. For that reason, this local Dockerfile
relies on Go's official image.

### DIFF
--- a/build-compose.sh
+++ b/build-compose.sh
@@ -18,7 +18,9 @@ then
   fi
 fi
 
-if [[ ! $(docker images -q tsuru-build) ]]; then docker build -t tsuru-build -f Dockerfile.build .; fi;
+context=$(mktemp -d)
+docker build -t tsuru-build -f Dockerfile.build $context
+rmdir $context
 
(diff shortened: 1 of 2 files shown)


build-compose.sh
```diff
@@ -18,8 +18,9 @@ then
   fi
 fi
 
+if [[ ! $(docker images -q tsuru-build) ]]; then docker build -t tsuru-build -f Dockerfile.build .; fi;
```
cezarsa: What do you think about always running something like:
```
context=$(mkdir -d)
docker build -t tsuru-build -f Dockerfile.build $context
rmdir $context
```

This would ensure the build runs almost instantaneously if the image is already up to date but it would also rebuild the image if changes to Dockerfile.build were detected, instead of requiring the tsuru-build to be removed manually.

andrestc: Good idea.

PR #2039, opened 2018-05-07
provision/kubernetes: prevent terminating pods from showing up as app units

andrestc:


provision/kubernetes/provisioner.go
```diff
@@ -312,11 +315,20 @@ func (p *kubernetesProvisioner) podsToUnitsMultiple(client *ClusterClient, pods
...
-		}
+		})
 	}
 	return units, nil
 }
 
+// merged from https://github.com/kubernetes/kubernetes/blob/1f69c34478800e150acd022f6313a15e1cb7a97c/pkg/quota/evaluator/core/pods.go#L333
+// and https://github.com/kubernetes/kubernetes/blob/560e15fb9acee4b8391afbc21fc3aea7b771e2c4/pkg/printers/internalversion/printers.go#L606
+func isTerminating(pod apiv1.Pod) bool {
+	if pod.Spec.ActiveDeadlineSeconds != nil && *pod.Spec.ActiveDeadlineSeconds >= int64(0) || pod.DeletionTimestamp != nil {
```
ggarnier: You could replace this `if` with `return pod.Spec.ActiveDeadlineSeconds != nil && *pod.Spec.ActiveDeadlineSeconds >= int64(0) || pod.DeletionTimestamp != nil`

Issue #1932, opened 2018-02-03
Kubernetes Cluster - NameSpace support for App creation.


harshapic:
Right now on kubernetes cluster there is only support for namespace @ cluster level, but no way to specify namespace while creation of app. Is it possible to provide support for namespace support @ app creation level

andrestc:
This is something we plan to add in the future, but probably as one namespace per pool.

PR #2043, opened 2018-05-10
Kubernetes pool namespaces

ggarnier:
Add a new configuration key, `kubernetes:use-pool-namespaces`. When set to true, the Kubernetes components created for each app (pods, services, deployments, etc) will use a different namespace per pool.

We still need to handle pool changing when this config is enabled, but I'm leaving it to another PR.

Refs #1932


provision/kubernetes/node.go
```diff
@@ -74,10 +75,25 @@ func (n *kubernetesNodeWrapper) ExtraData() map[string]string {
 }
 
 func (n *kubernetesNodeWrapper) Units() ([]provision.Unit, error) {
-	pods, err := appPodsFromNode(n.cluster, n.node.Name)
+	pools, err := pool.ListAllPools()
```
cezarsa: Do we need to list all pools here since the node belong to a single pool?

provision/kubernetes/provisioner.go
```diff
@@ -404,13 +404,21 @@ func podsForApps(client *ClusterClient, apps []provision.App) ([]apiv1.Pod, erro
 		}
 		sel = sel.Add(*req)
 	}
-	pods, err := client.CoreV1().Pods(client.Namespace()).List(metav1.ListOptions{
-		LabelSelector: sel.String(),
-	})
-	if err != nil {
-		return nil, err
+	poolNames := make([]string, len(apps))
```
cezarsa: After messing with kubectl a bit I found that it's possible to list pods from all namespaces with a single request, I think this should be used here, the only catch is that apparently we have to create the request more manually, something like: 
```
restCli, err := rest.RESTClientFor(client.restConfig)
var podList apiv1.PodList
err = restCli.Get().
	Resource("pods").
	VersionedParams(&metav1.ListOptions{
<opts>
}, scheme.ParameterCodec).
	Do().
	Into(&podList)
```

ggarnier: The problem is that this would list pods from all namespaces, not only the ones we use for pools - which include `kube-system`. This could be dangerous

provision/kubernetes/node.go
ggarnier: We need the namespace to list pods with a specific selector in https://github.com/ggarnier/tsuru/blob/8a066b8e3d34f6f5a6836c3962bbd42f04fa7989/provision/kubernetes/helpers.go#L413.

But with this alternative you just found out to list pods from all namespaces could replace this.

PR #2045, opened 2018-05-11
Improve kubernetes healthcheck compat. with old provisioner

cezarsa:
The following new fields were added to healthchecks in tsuru.yaml:

`force_restart`: Force pod restart on healthcheck failure (setting the
liveness probe)
`interval_seconds`: Interval between each healthcheck call in probes
`timeout_seconds`: Timeout for the healthcheck call.

For kubernetes provisioner:

Readiness probe now is only set if `use_in_router` flag is set to true,
otherwise we create the healthcheck call as a postStart lifecycle hook to
ensure it only runs once when the container is started.

For docker and swarm provisioners:

We now use the timeout_seconds parameter to control how long we wait
for the healthcheck call.


provision/kubernetes/deploy.go
```diff
@@ -306,29 +307,67 @@ func extraRegisterCmds(a provision.App) string {
...
+	if !hc.UseInRouter {
+		url := fmt.Sprintf("%s://localhost:%d/%s", hc.Scheme, port, strings.TrimPrefix(hc.Path, "/"))
+		// Use postStart lifecycle hook executing curl in a while loop. We do
+		// this because we only want this healthcheck to be called once the
+		// container is started however we must still retry it on connection
+		// errors or after a HTTP error if AllowedFailures is > 1. This is the
+		// reason why we only increment the counter if curl returns status code
+		// 22 (response code >= 400 according to curl docs)
+		result.postStartCmd = fmt.Sprintf(`i=0
+while [ $i -lt %[1]d ] && ! curl -sSf -X%[2]s -o /dev/null -m %[3]d %[4]s; do [ $? -eq 22 ] && let i=i+1; sleep 3; done
```
andrestc: `sleep 3` for compatibility with docker-cluster?

cezarsa: Yep, this was for compatibility as we have 3 seconds hardcoded in provision/docker. Using interval seconds here didn't seem right as the meaning is quite different from when it's used in the probes.

PR #2054, opened 2018-05-17
provision: ensure node removal removes healer data for node

cezarsa:
This PR moves existing node logic from the provision pkg to a newly created node pkg. Now, instead of handling healer data removal from inside the provisioner this logic was added to a RemoveNode method in the node pkg.


andrestc:
LGTM. I re-triggered the tests because it failed randomly.

PR #2056, opened 2018-05-18
Add Cluster specification to swagger api

jonnyguio:
Refs #1993

### DIFF
--- a/docs/reference/api.yaml
+++ b/docs/reference/api.yaml
@@ -1594,6 +1594,113 @@ paths:
         - pool
       security:
         - Bearer: []
+  /1.4/provisioners/clusters:
+    get:
+      operationId: ClusterList
+      description: List cluster
+      produces:
+      - application/json
+      responses:
+        '200':
+          description: Cluster
+          schema: 
+            type: array
+            items:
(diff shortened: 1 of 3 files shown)


docs/reference/api.yaml
```diff
@@ -2567,3 +2674,73 @@ definitions:
...
+      cacert:
+        type: string
+      clientcert:
+        type: string
+      pools:
+        type: array
+        items:
+          type: object
+          $ref: '#/definitions/Pool'
+      custom_data:
```
cezarsa: I think `custom_data` and `create_data` must be defined as:
```
custom_data:
  type: object
  additionalProperties:
    type: string
```

docs/reference/api.yaml
```diff
@@ -1357,6 +1357,113 @@ paths:
...
+        - Bearer: []
+    put:
+      operationId: ClusterUpdate
+      description: Update cluster.
+      parameters:
+        - name: ClusterUpdateData
+          in: body
+          required: true
+          schema:
+            $ref: '#/definitions/ClusterUpdateData'
```
cezarsa: The same `Cluster` definition can be used in the update call.

PR #2077, opened 2018-06-13
Using stderr for warning messages on server

gfleury:
Shell scripts snippets like 
```
token=$(tsurud token)
```

are broken due this msgs being print to stdout.


andrestc:
Thanks!

PR #2079, opened 2018-06-19
provision/kubernetes: no providerID in nodes, cloud controller should set it

cezarsa:


andrestc:
@cezarsa can we merge this?

PR #2080, opened 2018-06-19
docs: add documentation for event webhooks

ggarnier:


docs/managing/event-webhooks.rst
```diff
@@ -0,0 +1,77 @@
...
+Calling a specific URL every time a specific app is updated:
+
+.. highlight:: bash
+
+::
+
+    $ tsuru event-webhook-create my-webhook <my-url>
+            -t <my-team>
+            --kind-name app.update
+            --kind.value <my-app>
```
cezarsa: Should be `--target-value` here.

Issue #1735, opened 2017-09-15
Kubernetes: remove old deploy pods


ggarnier:
If some error occurs on an app deploy in a Kubernetes pool, a deploy or build pod (like `myapp-v1-deploy`) may be left behind. We should remove these pods before the next deploy. Here's an example of unit list (from `app-info` command) after a couple of unsuccessful deploys:

```
Units: 10
+------------------+---------+------+------+
| Unit             | Status  | Host | Port |
+------------------+---------+------+------+
| myapp-v10-build  | stopped |      |      |
| myapp-v10-deploy | error   |      |      |
| myapp-v11-build  | stopped |      |      |
| myapp-v11-deploy | error   |      |      |
| myapp-v12-build  | stopped |      |      |
| myapp-v12-deploy | stopped |      |      |
| myapp-v13-deploy | stopped |      |      |
| myapp-v14-build  | started |      |      |
| myapp-v14-deploy | stopped |      |      |
| myapp-v9-build   | stopped |      |      |
+------------------+---------+------+------+
```

Besides leaving this garbage, if the error happens during the build phase, the build pod is kept running (like `myapp-v14-build` above), which makes it impossible to make another deploy: `pods "myapp-v14-build" already exists`. Currently, the only way to solve this is manually deleting the pod.

VitorFalcao:
@ggarnier Hi, I would like to start contributing to the project. Could you please elaborate on the issue?

ggarnier:
@VitorFalcao I still didn't investigate on this issue. I just noticed that, after deploying an app a couple of times in my local environment, a few old deploy pods weren't removed. I'd have to check the `provisioner/kubernetes` package to help understand what could be causing this.

gfleury:
@ggarnier I was facing the same issue. In my situation was the lack of permissions for the Tsuru service account (for deleting pods). If you were using RBAC, it could be the same. 
The error was being hidden because cleanupPod() is only called by defer, so I guess that err is just being ignored. I'm adding a logging into cleanupPod just to make it more explicit by log when the deletion fails. 

https://github.com/gfleury/tsuru-helm-chart/commit/a7ab1e44b8af4800e4aa5db7a2c1ea87e44a1694

PR #2084, opened 2018-07-02
Adding logging to cleanupPod, as it only called from defer.

gfleury:
Doesn't solve #1735, but addresses it a bit since the lack of information on why the cleanUp failed on deleting the pods make the issue a bit more obscure.


andrestc:
LGTM, thanks!

PR #2091, opened 2018-07-13
service/broker: Uses UUIDs for service broker operations

andrestc:
Service Brokers requires that GUIDs and IDs are actual v4 UUIDs.
This PR creates a UUID field that is lazily generated for apps,
generates UUIDs for service instances and also stores broker instances
data inside the service instance data structure.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -17,6 +17,7 @@ import (
 
 	"github.com/globalsign/mgo"
 	"github.com/globalsign/mgo/bson"
+	"github.com/nu7hatch/gouuid"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/tsuru/tsuru/action"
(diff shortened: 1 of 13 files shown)


app/app.go
```diff
@@ -1333,6 +1337,29 @@ func (app *App) GetName() string {
...
+// if it does not exist.
+func (app *App) GetUUID() (string, error) {
+	if app.UUID != "" {
+		return app.UUID, nil
+	}
+	uuidV4, err := uuid.NewV4()
+	if err != nil {
+		return "", errors.WithMessage(err, "failed to generate uuid v4")
+	}
+	app.UUID = uuidV4.String()
```
cezarsa: By setting `app.UUID` here future calls to `GetUUID()` will always return it even if an error occurred while storing it on the database and this value won't ever be stored. It would be safer to only set `app.UUID` after a successful call to DB Update.

andrestc: nice catch. updated

PR #2092, opened 2018-07-16
Defining ResourceCPU from CpuShare() on app deploy spec

gfleury:
Add ResourceCPU on Kubernetes deployment spec. 
A plan with cpushare of 200 will result in 200m or 0.2.

### DIFF
--- a/provision/kubernetes/deploy.go
+++ b/provision/kubernetes/deploy.go
@@ -525,6 +525,11 @@ func createAppDeployment(client *ClusterClient, oldDeployment *v1beta2.Deploymen
 		resourceLimits[apiv1.ResourceMemory] = *resource.NewQuantity(memory, resource.BinarySI)
 		resourceRequests[apiv1.ResourceMemory] = *resource.NewQuantity(memory/overcommit, resource.BinarySI)
 	}
+	cpu := int64(a.GetCpuShare())
+	if cpu != 0 {
+		resourceLimits[apiv1.ResourceCPU] = *resource.NewMilliQuantity(cpu, resource.BinarySI)
+		resourceRequests[apiv1.ResourceCPU] = *resource.NewMilliQuantity(cpu/overcommit, resource.BinarySI)
+	}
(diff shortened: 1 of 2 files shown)


andrestc:
I'm a bit worried about compatibility issues with docker-cluster. Currently, the `CpuShare` field on docker pools is used to configure the `--cpu-shares` on docker. I think using ResourceCPU from the `CpuShare` plan field would make plans work either on docker or on Kubernetes and never on both. 

On docker, we use the `CpuShare` to set the `--cpu-shares` flag:

```
Set this flag to a value greater or less than the default of 1024 to increase or reduce the container’s weight, and give it access to a greater or lesser proportion of the host machine’s CPU cycles. This is only enforced when CPU cycles are constrained. When plenty of CPU cycles are available, all containers use as much CPU as they need. In that way, this is a soft limit. --cpu-shares does not prevent containers from being scheduled in swarm mode. It prioritizes container CPU resources for the available CPU cycles. It does not guarantee or reserve any specific CPU access.
```
I think the best approach would be to add a new field to plans, to be used as ResourceCPU on Kubernetes and as `--cpus` flag on docker:

```
Specify how much of the available CPU resources a container can use. For instance, if the host machine has two CPUs and you set --cpus="1.5", the container is guaranteed at most one and a half of the CPUs. This is the equivalent of setting --cpu-period="100000" and --cpu-quota="150000". Available in Docker 1.13 and higher.
```

But I would like others to also give opnions on this.

gfleury:
Hmmm, but do we have this amount of mixed environments, with docker and kubernetes on same Tsuru using same plans? If yes then it's better to implement a new field on Plan, otherwise documentation should be enough.

andrestc:
I see some mixed environments especially with users that want to start and try using tsuru with Kubernetes before committing to it. Whats your take on this, @cezarsa ?

gfleury:
Ok, I guess that creating an specific field for k8s cpushares should be the best approach, I will take a look on that.

Issue #2088, opened 2018-07-04
Handling token while having multiple targets on tsuru-client


gfleury:
Hi, 

So, what are the best options to handle tokens for multiple targets on the client?

Is there any specific reason that the token file doesn't have one token per line like?

target1 token1
target2 token2

Plugins are injecting token/target before running, don't know if it would work. I'm also looking for something more plug'n'play like tsuru --target, or tsuru -t. 

Any insights on that?

cezarsa:
I was thinking on something similar last week. Rather than multiple tokens in a single file I think it would be easier to manage having multiple token files, one for each target in `.tsuru/token.d/<target name>`.

To ease the migration process tsuru client could be changed to always look for the old `.tsuru/token` file if `.tsuru/token.d/<target name>` does not exist.

The second part of this change (changing the target for a single command or for a single shell instance) could be accomplished by changing how we handle the TSURU_TARGET environment variable.
Currently this environment variable must be a URL, we could change it to allow a target name. When selecting a target, tsuru would first list the existing targets from `.tsuru/targets` and if a target with name `$TSURU_TARGET` exists we would use it. If none exists we would fallback to the old behavior of using `$TSURU_TARGET` as a URL.

gfleury:
Hi, 
First part, having it separated in directories, looking by how it is implemented now, is easier to do this way. 
Second part solve the problem and it's declarative, but can mess the head of guys that don't have a shell background, because then it's a mix of setting env vars and parameters. 

$ TSURU_TARGET=target1 tsuru app-log -a app1

This is very helpful for using it embedded on CI jobs/scripts or something similar, for daily use it is safe, can prevent acting on the wrong target. But I would say that something similar to, 

$ tsuru -target target1 app-log -a app1

is more easy to use and remember on a daily basis.

cezarsa:
I agree, I think it would be nice to support `--target` argument in the tsuru command.

gfleury:
So seems that we have the token management a bit spread among tsuru, tsuru-client and go-tsuruclient now.. ReadToken is implemented on all three places and writetoken is implemented on tsuru/tsuru. Is there some guideline on how to do it? is go-tsuruclient going to be the shared package between all other packages?

cezarsa:
[...]

PR #2093, opened 2018-07-16
Creating token per target

gfleury:
Just POC with a token.d as suggested in #2088

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -8,6 +8,7 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net/http"
 	"os"
 	"os/signal"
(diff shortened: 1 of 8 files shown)


cmd/cmd.go
```diff
@@ -180,6 +184,36 @@ func (m *Manager) Run(args []string) {
 	} else if displayVersion {
 		args = []string{"version"}
 	}
+	if len(target) > 0 {
+		var targets = map[string]string{}
```
cezarsa: I don't think this logic should be here. Here you could simply call `os.Setenv("TSURU_TARGET", target)` and move this logic to the `ReadTarget()` using the `TSURU_TARGET` env as the `target` variable here.

Also, instead of manually parsing the targets file you can call the `getTargets()` function there which already returns a `map[string]string` for all targets in the targets file.

cezarsa:
Thank you, I'll merge this in the tsuru client later.

PR #2128, opened 2018-08-20
Service broker catalog cache

ggarnier:
In this PR, I've created a new service to handle caching service broker catalogs. The default cache time is 1 hour, but you can customize it per broker, with the `Config.CacheExpiration` flag in the api.

I also wanted to merge `ServiceBrokerService` and `ServiceBrokerCatalogCacheService` in a single service, but this PR is already too large.

### DIFF
--- a/api/service_broker.go
+++ b/api/service_broker.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"time"
 
 	"github.com/ajg/form"
 	"github.com/tsuru/tsuru/auth"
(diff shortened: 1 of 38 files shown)


api/service_broker.go
```diff
@@ -154,3 +143,27 @@ func serviceBrokerDelete(w http.ResponseWriter, r *http.Request, t auth.Token) e
...
+	dec := form.NewDecoder(nil)
+	dec.IgnoreCase(true)
+	dec.IgnoreUnknownKeys(true)
+	if err := request.ParseForm(); err != nil {
+		return nil, fmt.Errorf("unable to parse form: %v", err)
+	}
+	if err := dec.DecodeValues(&broker, request.Form); err != nil {
+		return nil, fmt.Errorf("unable to parse broker: %v", err)
+	}
+	cacheStr := request.FormValue("Config.CacheExpiration")
```
andrestc: Maybe we could use custom decoding for the service.Broker type?
https://github.com/ajg/form#custom-marshaling

ggarnier: I tried, but the support to custom marshaling is limited. I'd have to implement `encoding.TextUnmarshaler` in the `CacheExpiration` field itself, and it also doesn't allow setting the field to nil (which I do when this field is omitted)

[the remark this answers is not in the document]
ggarnier: Yes. We could make a PR to `ajg/form` to add support to `time.Duration` fields, but the project seems abandoned (the latest commit was 2 years ago).

Issue #2036, opened 2018-05-04
Improve platforms management


ggarnier:
We could make a lot of improvements in the way we manage platforms:

- add versioning
- allow rollbacks
- allow an app to lock on a specific platform version

andrestc:
Also https://github.com/tsuru/tsuru/issues/1530

PR #2131, opened 2018-08-22
api/platforms: add rollback endpoint

guilhermebr:
https://github.com/tsuru/tsuru/issues/2036

### DIFF
--- a/api/platform.go
+++ b/api/platform.go
@@ -235,3 +235,60 @@ func platformInfo(w http.ResponseWriter, r *http.Request, t auth.Token) error {
 	w.Header().Set("Content-Type", "application/json")
 	return json.NewEncoder(w).Encode(msg)
 }
+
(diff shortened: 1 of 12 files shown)


api/platform.go
```diff
@@ -235,3 +235,60 @@ func platformInfo(w http.ResponseWriter, r *http.Request, t auth.Token) error {
...
+// path: /platforms/{name}/rollback
+// method: POST
+// produce: application/x-json-stream
+// responses:
+//   200: OK
+//   400: BadRequest
+//   401: Unauthorized
+//   404: Not found
+func platformRollback(w http.ResponseWriter, r *http.Request, t auth.Token) (err error) {
+	r.ParseForm()
```
andrestc: I think we should start checking the error on every `r.ParseForm()` call.

PR #2154, opened 2018-09-06
provision/kubernetes: add clusterForPoolOrAny

### LINKED ISSUE #2153: Platform add fails when only a Kubernetes pool is available
It is reporting that there are no clusters, but there is one (and I`m able to deploy to it using a image deploy)
```
$ tsuru platform add static

Error: multiple errors reported (2): error 0: There is no docker node. 
Please list one in tsuru.conf or add one with `tsuru node-add`. - error 1: no cluster
```

guilhermebr:
fix #2153


guilhermebr:
@cezarsa PTAL

PR #2156, opened 2018-09-10
Improve name validation for Kubernetes

### LINKED ISSUE #2145: Fix name handling for Kubernetes
When an app runs on Kubernetes, we create a deployment for each app process. The deployment name is set as `<appname>-<processname>`, which can't have no more than 63 characters. Label names and values also can't have more than 63 characters.

The app name alone could have 63 characters, and the process names have no validation for length or acceptable characters.

ggarnier:
Change name validation rules to avoid incompatibilities with Kubernetes:

- reduce max length to 40 characters for app, platform, team token id, event webhook, pool, cluster, volume and service
- remove `tsuru-app-` prefix from `app` label in deployments and pods, to save some characters (label values can't have more than 63 characters)
- limit service and deployment names to 55 characters (max length allowed by Kubernetes is 63 characters). If `<app name>-<process name>`has more than 55 characters, replace the process name with a sha256 hash

This fixes #2145.

### DIFF
--- a/api/app_test.go
+++ b/api/app_test.go
@@ -1393,7 +1393,7 @@ func (s *S) TestCreateAppInvalidName(c *check.C) {
 	request.Header.Set("Authorization", "b "+token.GetValue())
 	s.testServer.ServeHTTP(recorder, request)
 	c.Assert(recorder.Code, check.Equals, http.StatusBadRequest)
-	msg := "Invalid app name, your app should have at most 63 " +
+	msg := "Invalid app name, your app should have at most 40 " +
 		"characters, containing only lower case letters, numbers " +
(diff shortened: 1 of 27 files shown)


ggarnier:
I just found out a couple of adjustments I need to make:

- [ ] app, cluster, volume: updates shouldn't validate names. Currently it would be impossible to update an app a name longer than 40 characters
- [ ] event webhook: improve the error message for invalid name (`Error: Status: 400 Bad Request, Body: name does not match regex "^[a-z][a-z0-9-]{0,39}$"`)

ggarnier:
@cezarsa PTAL

PR #2166, opened 2018-09-24
app/image: always return a existing process on GetImageWebProcessName

### LINKED ISSUE #2165: Routes rebuild infinite loop in Kubernetes
If an application has multiple processes and none of those processes are called `web`, the routes rebuild task is going to fail.

This happens because `GetImageWebProcessName` returns `web` and the Kubernetes provisioner is going to look for a service called <app>-web, resulting in an error that will cause the process to start over.

andrestc:
Prior to this change, `GetImageWebProcessName` would return "web" if the
application had more than a single process, even if none of them was
actually called "web". This causes a issue on the Kubernetes provisioner
because it relies on the returned name to find some cluster resources,
causing the `RoutableAddresses()` method to fail.

This change makes sure that `GetImageWebProcessName` always return a
existing process name. If the application has more than a single process
it will either return "web", if one of them is called "web", or the
first process in alphabetical order.

Fixes #2165

### DIFF
--- a/app/image/image.go
+++ b/app/image/image.go
@@ -7,6 +7,7 @@ package image
 import (
 	"fmt"
 	"regexp"
+	"sort"
 	"strings"
 
 	"github.com/globalsign/mgo"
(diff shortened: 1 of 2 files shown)


app/image/image.go
```diff
@@ -168,20 +169,23 @@ func GetImageMetaData(imageName string) (ImageMetadata, error) {
 }
 
 func GetImageWebProcessName(imageName string) (string, error) {
-	processName := "web"
+	const processName = "web"
 	data, err := GetImageMetaData(imageName)
 	if err != nil {
 		return processName, err
```
andrestc: Perhaps we should return "" here?

PR #2175, opened 2018-10-03
app: validate volumes when changing pools

ggarnier:
Apps with binded volumes can't be migrated between different provisioners.


cezarsa:
If both old and new provisioners implement the `VolumeProvisioner` interface, wouldn't it be safe to allow the update to run?

andrestc:
I think we could merge this as is for a 1.6.x release, since the update is broken atm and work on implementing the volume migration in the future.

Issue #2138, opened 2018-08-28
Add "simplified" flag to app list


ggarnier:
`app list` command in tsuru client supports a `-q` flag to return only the app names, but it's implemented in the client. We could add this flag to the api, making this call much faster. The dashboard initial page would benefit from this.

aman12592:
@ggarnier I would like to contribute towards resolving the issue. If you could please provide any points on getting started would be of great help.

ggarnier:
@aman12592 the idea is adding support to a new flag in `/apps` endpoint (something like `nameOnly=true`). If this flag is present, we would return only the app names (no need to check units status).

The code for this is in `api/app.go` file. Let us know if you need more help.

juanjcsr:
Hi! I am working on this issue. Is there any preferred format for the output?

I can return the names in a simple array like this:
```
[
   "app-name1",
   "app-name2"
]
``` 

Or the output could be more complex, such as:

```
[
    {"name": "app-name1"},
    {"name": "app-name2"}
]
```

Or, if there is another output format, I can make it happen :smile:

ggarnier:
@juanjcsr I prefer the second option, because it keeps the contract of the API:

```
[
    {"name": "app-name1"},
    {"name": "app-name2"}
]
```

cezarsa:
Closed by #2177

PR #2177, opened 2018-10-08
api: add nameOnly flag to apps endpoint

juanjcsr:
This is my attempt to fix issue #2138 

I added a new flag: `nameOnly`. If the flag is present and true, then it returns a JSON with an array of app names in the form: 
```
 [
    {"name": "app-name1"},
    {"name": "app-name2"}
]
```
It also skips the query for the app unit status.

I also added one test case to check the flag does not break the current interface implementation.

### DIFF
--- a/api/app.go
+++ b/api/app.go
@@ -222,6 +222,9 @@ func appList(w http.ResponseWriter, r *http.Request, t auth.Token) error {
 		for i, ap := range apps {
 			ur := app.AppUnitsResponse{Units: nil, Err: nil}
 			miniApps[i], err = minifyApp(ap, ur)
+			if err != nil {
+				return err
+			}
 		}
 		return json.NewEncoder(w).Encode(miniApps)
 	}
--- a/api/app.go
+++ b/api/app.go
@@ -219,18 +219,9 @@ func appList(w http.ResponseWriter, r *http.Request, t auth.Token) error {
 	w.Header().Set("Content-Type", "application/json")
 	miniApps := make([]miniApp, len(apps))
 	if simple {
-		for i, app := range apps {
-			ma := miniApp{
-				Name:      app.Name,
-				Pool:      app.Pool,
-				Plan:      app.Plan,
-				TeamOwner: app.TeamOwner,
-				CName:     app.CName,
-				Routers:   app.Routers,
-				Lock:      &app.Lock,
-				Tags:      app.Tags,
-			}
-			miniApps[i] = ma
+		for i, ap := range apps {
+			ur := app.AppUnitsResponse{Units: nil, Err: nil}
+			miniApps[i], err = minifyApp(ap, ur)
 		}
 		return json.NewEncoder(w).Encode(miniApps)
 	}
@@ -238,8 +229,6 @@ func appList(w http.ResponseWriter, r *http.Request, t auth.Token) error {
 	if err != nil {
 		return err
 	}
-
-	// miniApps := make([]miniApp, len(apps))
 	for i, app := range apps {
 		miniApps[i], err = minifyApp(app, appUnits[app.Name])
 		if err != nil {
--- a/api/app.go
+++ b/api/app.go
@@ -215,11 +215,24 @@ func appList(w http.ResponseWriter, r *http.Request, t auth.Token) error {
 		w.WriteHeader(http.StatusNoContent)
 		return nil
 	}
+	w.Header().Set("Content-Type", "application/json")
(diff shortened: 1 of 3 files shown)


api/app_test.go
```diff
@@ -436,6 +436,54 @@ func (s *S) TestAppListFilteringByStatusIgnoresInvalidValues(c *check.C) {
...
+	request.Header.Set("Authorization", "b "+s.token.GetValue())
+	recorder := httptest.NewRecorder()
+	s.testServer.ServeHTTP(recorder, request)
+	c.Assert(recorder.Code, check.Equals, http.StatusOK)
+	c.Assert(recorder.Header().Get("Content-Type"), check.Equals, "application/json")
+	var apps []app.App
+	err = json.Unmarshal(recorder.Body.Bytes(), &apps)
+	c.Assert(err, check.IsNil)
+	c.Assert(apps, check.HasLen, 2)
+	c.Assert(apps[0].Name, check.Equals, app1.Name)
```
andrestc: I think it would be better here if we tested that *only* the app name is set, e.g:
`c.Assert(apps[0], check.Equals, App{Name: "app1})`

api/app.go
```diff
@@ -215,11 +215,24 @@ func appList(w http.ResponseWriter, r *http.Request, t auth.Token) error {
 		w.WriteHeader(http.StatusNoContent)
 		return nil
 	}
+	w.Header().Set("Content-Type", "application/json")
+	simple, _ := strconv.ParseBool(r.URL.Query().Get("nameOnly"))
+	if simple {
+		appsNames := make([]interface{}, len(apps))
```
andrestc: Instead of using a `[]interface{}`, I think it is better to use a `[]miniApp` here and then set each miniApp's name on the loop.

juanjcsr: I agree. In fact, that was my first attempt. However, since `miniApp` struct fields implement json decoder without  `omitempty`, I ended with the following JSON which I think it is not correct since it implies the App has all its fields empty.

```
[{
    "name": "App1",
    "pool": "",
    "teamowner": "",
    "plan": {
        "name": "",
        "memory": 0,
        "swap": 0,
        "cpushare": 0
    },
    "units": null,
    "cname": null,
    "ip": "",
    "routers": null,
    "lock": null,
    "tags": null
}]
```


In order to try and test another attempt, I added `omitempty`to `miniApp`  and ended with a result much closer to the expected solution: 

```
[{"name":"App1","plan":{"name":"","memory":0,"swap":0,"cpushare":0}}]
```

However, I feel this is not quite right since I had to modify the whole miniApp json struct just for this issue. That is why I went with `[]interface{}`

I looked around but almost every solution used reflection to remove the empty struct fields on the fly which I think its an overhead... What do you think?

andrestc: Hi @juanjcsr, thanks for getting back. I was thinking about this issue one more time and I think we could simplify (pun not intended) our approach:

Instead of using `nameOnly` as a query flag, we could use `simplified` then simply return the entire `miniApp` without the units (which is the complex part). You could do that just by skipping the call to `app.Units` if this flag is set.

What do you think?

juanjcsr: I think this is a way better solution. It solves the original issue while still being fast by skipping the Units part. 

I will upload my changes latter today and will also add documentation to it.

api/app_test.go
juanjcsr: Hi! I updated the test case. It now checks for an empty `units` array.

api/app.go
juanjcsr: Hi @andrestc, I updated my code to support a new `simplified` flag that returns an array of `miniApp` without the `app.Units` part.

api/app.go
```diff
@@ -215,25 +215,31 @@ func appList(w http.ResponseWriter, r *http.Request, t auth.Token) error {
...
 	w.Header().Set("Content-Type", "application/json")
-	simple, _ := strconv.ParseBool(r.URL.Query().Get("nameOnly"))
+	miniApps := make([]miniApp, len(apps))
 	if simple {
-		appsNames := make([]interface{}, len(apps))
-		for i, appName := range apps {
-			res := map[string]interface{}{
-				"name": appName.Name,
+		for i, app := range apps {
+			ma := miniApp{
```
andrestc: I think we can use `minifyApp` here, passing a AppUnitsResponse with `.Error = nil` and `.Units = nil`.

juanjcsr:
Done :) Thank you for your input!

andrestc:
Thanks!

PR #2180, opened 2018-10-09
Kubernetes: apps with binded volumes can't change namespaces

ggarnier:
Currently we don't support moving volumes between namespaces of a Kubernetes cluster. This PR adds a validation to prevent this.

Moving between different clusters works fine.


andrestc:
LGTM.

PR #2189, opened 2018-10-18
kubernetes: add rebuild option

guilhermebr:


provision/kubernetes/builder.go
```diff
@@ -77,6 +78,32 @@ func (c *KubeClient) ImageTagPushAndInspect(a provision.App, imageID, newImage s
...
+		return nil, err
+	}
+	reader, writer := io.Pipe()
+	stderr := &bytes.Buffer{}
+	go func() {
+		opts := execOpts{
+			client: client,
+			app:    app,
+			image:  imageName,
+			cmds:   []string{"cat /home/application/archive.tar.gz"},
```
cezarsa: I don't think this will work as a shell is not automatically started for running the command, this should probably be:
```suggestion
			cmds:   []string{"cat", "/home/application/archive.tar.gz"},
```

provision/kubernetes/builder.go
```diff
@@ -77,6 +78,32 @@ func (c *KubeClient) ImageTagPushAndInspect(a provision.App, imageID, newImage s
...
+			stdout: writer,
+			stderr: stderr,
+		}
+		err := runIsolatedCmdPod(client, opts)
+		if err != nil {
+			writer.CloseWithError(err)
```
cezarsa: `stderr` is never used, maybe it could be used here as in:
```suggestion
			writer.CloseWithError(errors.Wrapf(err, "error reading archive, stderr: %q", stderr.String()))
```

guilhermebr:
@cezarsa PTAL

PR #2190, opened 2018-10-18
app: enable filtering apps with specific platform versions

andrestc:
Prior to this change, users could not filter apps that were using a
specific plataform version (locked).

This change makes it possible to filter apps that are locked to a
specific platform version or are not locked to a specific version. Usage
is as follows:

`app list -p go` -> returns all apps using go (locked or not)
`app list -p go:v1` -> returns all apps that are using go locked to the
v1 version
`app list -p go:latest` -> return all apps that are not locked to a
specific go platform version.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -1842,7 +1842,21 @@ func (f *Filter) Query() bson.M {
 		query["teamowner"] = f.TeamOwner
 	}
 	if f.Platform != "" {
-		query["framework"] = f.Platform
+		parts := strings.SplitN(f.Platform, ":", 2)
+		query["framework"] = parts[0]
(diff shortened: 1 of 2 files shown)


app/app.go
```diff
@@ -1842,7 +1842,21 @@ func (f *Filter) Query() bson.M {
 		query["teamowner"] = f.TeamOwner
 	}
 	if f.Platform != "" {
-		query["framework"] = f.Platform
+		parts := strings.SplitN(f.Platform, ":", 2)
```
cezarsa: I think you can use `image.SplitImageName` here, there are some edge cases from what I remember.

PR #2201, opened 2018-10-25
provision/kubernetes: adds a way to append labels/annotations as is

andrestc:
Prior to this change, every label/annotation added to a node thru the
metadata fields would contain the "tsuru.io/" prefix. This makes it hard
to use the metadata field as a way to add annotations and labels
required by some Kubernetes components.

This change introduces two special metadata fields:
"tsuru.io/extra-labels" and "tsuru.io/extra-annotations". Each of these
fields may contain multiple key and values that are applied as is as
labels ou annotations on a Kubernetes node.

### DIFF
--- a/provision/kubernetes/provisioner.go
+++ b/provision/kubernetes/provisioner.go
@@ -645,14 +645,12 @@ func setNodeMetadata(node *apiv1.Node, pool, iaasID string, meta map[string]stri
 			appendKV(v, ",", "=", node.Annotations)
 		case tsuruExtraLabelsMeta:
 			appendKV(v, ",", "=", node.Labels)
-		default:
-			if v == "" {
-				delete(node.Annotations, k)
-				continue
-			}
-			node.Annotations[k] = v
 		}
-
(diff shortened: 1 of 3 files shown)


provision/kubernetes/provisioner.go
```diff
@@ -640,11 +640,19 @@ func setNodeMetadata(node *apiv1.Node, pool, iaasID string, meta map[string]stri
...
-		if v == "" {
-			delete(node.Annotations, k)
-		} else {
+		switch k {
+		case tsuruExtraAnnotationsMeta:
+			appendKV(v, ",", "=", node.Annotations)
+		case tsuruExtraLabelsMeta:
+			appendKV(v, ",", "=", node.Labels)
+		default:
+			if v == "" {
```
cezarsa: Shouldn't the default section be actually outside the switch statement to ensure the annotation `tsuru.io/extra-labels,annotation` itself is added to the node?

andrestc:
PTAL @cezarsa

PR #2228, opened 2018-11-28
api/server: prevent to offer invalid TLS certificates

nettoclaudio:
This change adds a new flag option on Tsuru API configuration file to
prevent it offers invalids certificates to its clients. In worst case,
the web server is gracefully shutdown to avoid other problems.

I can think a couple of problems that can occur when the certificates
served by Tsuru API to expire. For instance, one of these is related
between Big Sibling and Tsuru API. If BS can't talk with Tsuru API
(because certificate became invalid), the last one will think the node
(related to that BS) died and will heal it. Now imagine this for
many nodes... That's may be a big problem. Because of that, I think
it's safer validate certificates on Tsuru API before to offer them.


api/server.go
```diff
@@ -755,15 +766,130 @@ func (cr *certificateReloader) start() {
...
+					continue
+				}
+				nextValidation := time.Until(certificate.NotAfter)
+				err = validateTLSCertificate(cv.conf.certificate, cv.conf.roots)
+				cv.conf.Unlock()
+				if err != nil {
+					log.Errorf("[certificate-validator] the currently loaded certificate is invalid: %v\n", err)
+					cv.shutdownServerFunc(err)
+				} else {
+					log.Debugf("[certificate-validator] certificate is valid, next validation scheduled to %s", nextValidation)
```
andrestc: I think this doesn't need to be set to `debug`, a regular Print looks ok.

docs/reference/config.rst
```diff
@@ -77,6 +77,17 @@ tls:key-file
...
 
+tls:validate-certificate
+++++++++++++++++++++++++
+
+``tls:validate-certificate`` prevents an invalid certificate from being offered 
+to web clients or not. This setting is optional and defaults to "false".
+
+If enabled, the server will validate the certificates before server's start and
+during the certificates auto-reload process (if any). If a certificate expires
+during server execution (after loaded by server), so this feature will gracefully
```
andrestc:
```suggestion
during server execution (after loaded by server) this feature will gracefully
```

api/server.go
```diff
@@ -753,15 +764,130 @@ func (cr *certificateReloader) start() {
 	})
 }
 
+type certificateValidator struct {
+	conf   *srvConfig
+	stopCh chan bool
```
andrestc:
```suggestion
	stopCh chan bool
	stopDoneCh chan bool
```

api/server.go
```diff
@@ -753,15 +764,130 @@ func (cr *certificateReloader) start() {
...
+	// certificates become invalid. If not defined, its default action is
+	// gracefully shutdown the server.
+	shutdownServerFunc func(error)
+}
+
+func (cv *certificateValidator) Shutdown(ctx context.Context) error {
+	if cv.stopCh == nil {
+		return nil
+	}
+	close(cv.stopCh)
```
andrestc:
```suggestion
	close(cv.stopCh)
	select {
	    case <-cv.stopDoneCh:
	        return nil
	    case <-ctx.Done:
	        return ctx.Err()
	 }
```

api/server.go
```diff
@@ -753,15 +764,130 @@ func (cr *certificateReloader) start() {
...
+				} else {
+					fmt.Printf("[certificate-validator] certificate is valid, next validation scheduled to %s\n", nextValidation)
+				}
+				log.Debug("[certificate-validator] finishing certificate validator")
+				select {
+				case <-time.After(nextValidation):
+				case <-cv.conf.certificateReloadedCh:
+					continue
+				case <-cv.stopCh:
+					return
```
andrestc:
```suggestion
                                        cv.stopDoneCh <- true
					return
```

PR #2232, opened 2018-12-04
db: add log message when can't create app log collection

ggarnier:
MongoDB collections for app logs are created as [capped collections](https://docs.mongodb.com/manual/core/capped-collections/). But if the collection creation fails for any reason, the error is ignored, and the first insert ends up creating the collection, but uncapped.

This PR adds a log message when that occurs.


ggarnier:
Closed in favor of #2236

PR #2245, opened 2019-01-11
cmd: wait for server shutdown before returning

cezarsa:
This ensures that the client won't exit before the http response was
completely flushed, as this would cause the browser to show a connection
error.


cezarsa:
@fsouza volta Chico!

PR #2248, opened 2019-01-14
api: Test for token-info

sacardoso:


api/team_token.go
```diff
@@ -38,6 +38,34 @@ func tokenList(w http.ResponseWriter, r *http.Request, t auth.Token) error {
 	return json.NewEncoder(w).Encode(tokens)
 }
 
+// title: token info
+// path: /tokens/{token_id}
+// method: GET
+// produce: application/json
+// responses:
+//   200: List token
```
cezarsa: should be `Get token`

api/team_token.go
```diff
@@ -38,6 +38,34 @@ func tokenList(w http.ResponseWriter, r *http.Request, t auth.Token) error {
...
+//   204: No content
```
cezarsa: There is no situation where this could return 204, this line can be removed.

api/team_token.go
```diff
@@ -38,6 +38,34 @@ func tokenList(w http.ResponseWriter, r *http.Request, t auth.Token) error {
...
+	if err != nil {
+		if err == authTypes.ErrTeamTokenNotFound {
+			return &errors.HTTP{
+				Code:    http.StatusNotFound,
+				Message: err.Error(),
+			}
+		}
+		return err
+	}
+	w.Header().Set("Content-Type", "application/json")
```
cezarsa: I'm sorry I forgot about this part when we were discussing this issue.
We need to check for permission to read the found token, for team tokens we have two different checks that need to be made, the first one is just like the other handlers checking if we can read the token:

```
allowed := permission.Check(t, permission.PermTeamTokenRead,
    permission.Context(permTypes.CtxTeam, teamToken.Team),
)
if !allowed {
}
```
The second is particular to team tokens, we only let the user see the token value if the token permissions is a subset of user permissions, if it's not we must set the token value to an empty string. The logic will be the same as in https://github.com/tsuru/tsuru/blob/a80bbcacaec00b4e268cfe1cc023456427c4bed3/auth/team_token.go#L229-L235

I think this logic can be extracted to a new public function and this handler will call it.

api/team_token_test.go
```diff
@@ -238,3 +238,26 @@ func (s *S) TestTeamTokenUpdate(c *check.C) {
...
+	c.Assert(err, check.IsNil)
+
+	request, err := http.NewRequest("GET", "/1.6/tokens/id1", nil)
+	c.Assert(err, check.IsNil)
+	request.Header.Set("Authorization", "bearer "+s.token.GetValue())
+	recorder := httptest.NewRecorder()
+	s.testServer.ServeHTTP(recorder, request)
+	c.Assert(recorder.Code, check.Equals, http.StatusOK)
+
+	token, err := servicemanager.TeamToken.FindByTokenID("id1")
```
ggarnier: You don't need to check the token here. You could use `newToken` returned by `servicemanager.TeamToken.Create`

Issue #2213, opened 2018-11-01
Kubernetes: extended ports configuration


andrestc:
It is a common request to be able to define multiple ports per processes. This issue extends the proposed yaml structure on #2212 to be able to specify different ports and also protocols to be used in Kubernetes.

Once again, the app has the following Procfile:

```
web: ...
sidecar: ...
mysql: ...
```

The user is able to configure different and multiple ports per process as:

```yaml
kubernetes:
  groups:
    web:
      web:
        ports:
          - name: apache-http # an optional name for the port 
            protocol: http
            port: 80 # The port that is going to be exposed on the router.
            target_port: 8888 # The port on which the application listens on.
          - name: apache-https
            protocol: https
            port: 443
            target_port: 4444
      sidecar:
        ports:
          - name: metrics-https
            protocol: http # port defaults to the protocol default port (when available)
    mysql:
      mysql:
        ports:
          - protocol: tcp # name defaults to <process>-<protocol>, mysql-tcp in this case
            port: 3306 # target_port defaults to the same value as port
```

This will require some changes on how tsuru interacts with routers. 
1. Each router needs a way to specify what protocols it suports
2. We could create some default domain names for some protocols, such as: <process>.<group>.<domain>. For example: `web.web.myrouter.cloud.com` and `sidecar.web.myrouter.cloud.com`. (Related to https://github.com/tsuru/tsuru/issues/1499)

ggarnier:
We could also support selecting which ports will be added as liveness/readiness probes.

ggarnier:
We could support disabling port expose for a specific process using an empty array:

```
kubernetes:
  groups:
    mypod:
      web:
        ports:
          - target_port: 8000
      worker:
        ports: []
```

Processes declared in `Procfile` and omitted in this config would expose the default port, and processes declared in this config with an empty array wouldn't expose any ports.

What do you think of this approach?

andrestc:
If `tsuru.yaml` has nothing configurated under `kubernetes:` we should keep the current behavior (first process or the process called web exposes the `8888`).
If `tsuru.yaml` has anything under `kubernetes:` I think its better if we require that every open port should be explicitly set (more secure and Its what I would expect). So the following yaml means that no port is being exposed on the worker process.

```
kubernetes:
  groups:
    mypod:
      web:
        ports:
          - target_port: 8000
      worker:
````

ggarnier:
[...]

PR #2249, opened 2019-01-14
Add support to multiple ports on Kubernetes

ggarnier:
This PR adds support to customize exposed ports for each process of an app on Kubernetes. Now we can expose more than one port for each process (both TCP and UDP), or config a process to avoid exposing any ports (like a worker).

This is a working implementation, although there are some limitations that I decided to leave to a future PR:

- healthcheck is set only on the first port of a process. Eventually we need to make this configurable
- the router port is attached to the first exposed port of the web process. We also need to discuss how to better handle this
- documentation is missing

Refs #2213

### DIFF
--- a/builder/kubernetes/builder.go
+++ b/builder/kubernetes/builder.go
@@ -70,9 +70,6 @@ func imageBuild(client provision.BuilderKubeClient, a provision.App, imageID str
 	if err != nil {
 		return "", err
 	}
-	if len(imageInspect.Config.ExposedPorts) > 1 {
-		return "", errors.Errorf("too many ports exposed in Dockerfile, only one allowed: %+v", imageInspect.Config.ExposedPorts)
-	}
 	procfile := image.GetProcessesFromProcfile(procfileRaw)
 	if len(procfile) == 0 {
 		fmt.Fprintln(evt, " ---> Procfile not found, using entrypoint and cmd")
@@ -90,8 +87,11 @@ func imageBuild(client provision.BuilderKubeClient, a provision.App, imageID str
 		Processes:  procfile,
 		CustomData: tsuruYamlToCustomData(tsuruYaml),
 	}
+	imageData.ExposedPorts = make([]string, len(imageInspect.Config.ExposedPorts))
+	i := 0
 	for k := range imageInspect.Config.ExposedPorts {
-		imageData.ExposedPort = string(k)
+		imageData.ExposedPorts[i] = string(k)
+		i++
 	}
 	err = imageData.Save()
 	if err != nil {
--- a/builder/kubernetes/builder.go
+++ b/builder/kubernetes/builder.go
@@ -104,10 +104,17 @@ func tsuruYamlToCustomData(yaml *provision.TsuruYamlData) map[string]interface{}
 	if yaml == nil {
 		return nil
 	}
-	return map[string]interface{}{
-		"healthcheck": yaml.Healthcheck,
-		"hooks":       yaml.Hooks,
+	customData := map[string]interface{}{}
+	if yaml.Healthcheck.Path != "" {
+		customData["healthcheck"] = yaml.Healthcheck
 	}
+	if len(yaml.Hooks.Build) != 0 || len(yaml.Hooks.Restart.Before) != 0 || len(yaml.Hooks.Restart.After) != 0 {
+		customData["hooks"] = yaml.Hooks
+	}
+	if len(yaml.Kubernetes.Groups) != 0 {
+		customData["kubernetes"] = yaml.Kubernetes
+	}
+	return customData
 }
 
 func downloadFromContainer(client provision.BuilderKubeClient, app provision.App, evt *event.Event) (io.ReadCloser, error) {
--- a/builder/kubernetes/builder.go
+++ b/builder/kubernetes/builder.go
@@ -104,17 +104,11 @@ func tsuruYamlToCustomData(yaml *provision.TsuruYamlData) map[string]interface{}
 	if yaml == nil {
 		return nil
 	}
-	customData := map[string]interface{}{}
-	if yaml.Healthcheck.Path != "" {
-		customData["healthcheck"] = yaml.Healthcheck
+	return map[string]interface{}{
+		"healthcheck": yaml.Healthcheck,
+		"hooks":       yaml.Hooks,
+		"kubernetes":  yaml.Kubernetes,
 	}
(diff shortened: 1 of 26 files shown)


provision/kubernetes/deploy.go
```diff
@@ -1030,6 +1047,53 @@ func (m *serviceManager) DeployService(ctx context.Context, a provision.App, pro
...
+			svcPorts[i].Port = int32(defaultPort)
+		}
+		if len(port.Name) > 0 {
+			svcPorts[i].Name = port.Name
+		} else {
+			svcPorts[i].Name = fmt.Sprintf("%s-%d", defaultHttpPortName, i+1)
+		}
+	}
+	if len(svcPorts) == 0 {
+		defaultPort, _ := strconv.Atoi(provision.WebProcessDefaultPort())
```
cezarsa: Redeclaration of `defaultPort`

provision/kubernetes/deploy.go
```diff
@@ -1030,6 +1047,53 @@ func (m *serviceManager) DeployService(ctx context.Context, a provision.App, pro
...
+		svcPorts = append(svcPorts, apiv1.ServicePort{
+			Protocol:   "TCP",
```
cezarsa: For consistency we could use the constant `apiv1.ProtocolTCP`.

provision/kubernetes/provisioner.go
```diff
@@ -1313,3 +1314,63 @@ func appCustomResourceDefinition() *v1beta1.CustomResourceDefinition {
...
+		return envs
+	}
+
+	yamlData, err := image.GetImageTsuruYamlData(imageName)
+	if err != nil {
+		return append(envs, provision.DefaultWebPortEnvs()...)
+	}
+	portsConfig, err := getProcessPortsForImage(imageName, yamlData, process)
+	if err != nil || len(portsConfig) == 0 {
+		return removeDefaultPortEnvs(envs)
```
cezarsa: Should the default env really be removed when loading an invalid config?

provision/provision.go
```diff
@@ -702,6 +703,23 @@ type TsuruYamlHealthcheck struct {
 	TimeoutSeconds  int    `json:"timeout_seconds" yaml:"timeout_seconds" bson:"timeout_seconds,omitempty"`
 }
 
+type TsuruYamlKubernetesConfig struct {
+	Groups map[string]TsuruYamlKubernetesGroup `bson:",omitempty"`
```
cezarsa: This will be a pain to handle... Both pod names and process names can have `.` in their names. Serializing this to mongodb as a map this will cause an error as keys aren't allowed to have dots. We would need to change the way we serialize this to mongodb to use a list with keys and values to correctly allow dots here. This will probably mean using different structs for yaml parsing and database serialization. :(

builder/kubernetes/builder.go
```diff
@@ -104,10 +104,17 @@ func tsuruYamlToCustomData(yaml *provision.TsuruYamlData) map[string]interface{}
...
 		return nil
 	}
-	return map[string]interface{}{
-		"healthcheck": yaml.Healthcheck,
-		"hooks":       yaml.Hooks,
+	customData := map[string]interface{}{}
+	if yaml.Healthcheck.Path != "" {
+		customData["healthcheck"] = yaml.Healthcheck
+	}
+	if len(yaml.Hooks.Build) != 0 || len(yaml.Hooks.Restart.Before) != 0 || len(yaml.Hooks.Restart.After) != 0 {
```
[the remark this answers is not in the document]
ggarnier: I added that to avoid saving an empty healthcheck structure for every app without HC. But you have a point

provision/kubernetes/deploy.go
```diff
@@ -590,6 +590,19 @@ func createAppDeployment(client *ClusterClient, oldDeployment *v1beta2.Deploymen
...
+	for i, port := range processPorts {
+		var portInt int
+		if port.TargetPort > 0 {
+			portInt = port.TargetPort
+		} else if port.Port > 0 {
+			portInt = port.Port
+		} else {
+			portInt, _ = strconv.Atoi(provision.WebProcessDefaultPort())
+		}
+		containerPorts[i].ContainerPort = int32(portInt)
```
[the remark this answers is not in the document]
ggarnier: I don't think it's worth. A duplicated port would lead to an error creating the deployment, and the error message clearly shows what's wrong, like in this example:

```
Service "multiport-web" is invalid: spec.ports[2]: Duplicate value: core.ServicePort{Name:"", Protocol:"TCP", Port:9999, TargetPort:intstr.IntOrString{Type:0, IntVal:0, StrVal:""}, NodePort:0}
```

provision/kubernetes/deploy.go
```diff
@@ -1030,6 +1047,53 @@ func (m *serviceManager) DeployService(ctx context.Context, a provision.App, pro
...
+	}
+	processPorts, err := getProcessPortsForImage(imgName, yamlData, processName)
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+	svcPorts := make([]apiv1.ServicePort, len(processPorts))
+	defaultPort, _ := strconv.Atoi(provision.WebProcessDefaultPort())
+	for i, port := range processPorts {
+		if len(port.Protocol) > 0 {
+			svcPorts[i].Protocol = apiv1.Protocol(port.Protocol)
```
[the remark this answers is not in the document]
ggarnier: Just like in duplicated ports, the error message is clear in this case:

```
Service "multiport-web" is invalid: spec.ports[0].protocol: Unsupported value: "FOO": supported values: "TCP", "UDP"
```

provision/kubernetes/provisioner.go
ggarnier: That makes sense. I'll change to remove the default envs only when `err == nil && len(portsConfig) == 0` (no ports exposed for this process).

provision/kubernetes/deploy.go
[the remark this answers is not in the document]
cezarsa: Imagine we have a the following configuraration: 

```
"kubernetes": {
    "groups": {
        "pod1": {
            "web": {
                "ports": [
                    {
                        "port": 1080,
                        "target_port": 8000
                    },
                    {
                        "port": 2080,
                        "target_port": 8000
                    }
                ]
            }
        }
    }
}
```

It is valid, but it would fail here because we would have 2 entries setting `ContainerPort` to `8000`. In this situation having a single port entry would be the right thing to do.

Issue #2230, opened 2018-11-28
Kubernetes: External Traffic Policy


guilhermebr:
Its desirable to have a way to choose per cluster the external traffic policy for services, that by default is `Cluster`.

Choosing `Local` as the policy have benefits like described here: https://www.asykim.com/blog/deep-dive-into-kubernetes-external-traffic-policies

For `Local` policy we need a way to keep routers updated with only Node addresses that have Pods for the app.

PR #2251, opened 2019-01-15
Configuration for traffic policy local and local routable addresses

cezarsa:
This PR introduces 2 new flags to the kubernetes cluster.

`external-policy-local` will set the external policy local on created services. Please beware that setting this to true may cause flapping or some downtime during deploys when the app router uses node ips/ports.

`router-local` will only add to the router node ips where there are ready pods for the app. This works independenly from the chosen traffic policy. Setting this to true is safe when using the default Cluster traffic policy and should reduce the number of backends on the router.


provision/kubernetes/monitor.go
```diff
@@ -0,0 +1,125 @@
+// Copyright 2016 tsuru authors. All rights reserved.
```
andrestc: 2019

provision/kubernetes/cluster.go
```diff
@@ -135,11 +137,12 @@ func NewClusterClient(clust *provTypes.Cluster) (*ClusterClient, error) {
 	if err != nil {
 		return nil, err
 	}
-	return &ClusterClient{
+	clusterClient := &ClusterClient{
```
ggarnier: What's the point of this change?

provision/kubernetes/helpers.go
```diff
@@ -784,3 +784,19 @@ func waitForContainerFinished(ctx context.Context, client *ClusterClient, podNam
 		return false, nil
 	}, nil)
 }
+
+func isPodReady(pod *apiv1.Pod) bool {
+	for _, cond := range pod.Status.Conditions {
+		if cond.Type == apiv1.PodReady {
```
ggarnier: You could join these 2 `if`s into one.

provision/kubernetes/monitor.go
```diff
@@ -0,0 +1,125 @@
...
+	if newPod.ResourceVersion == oldPod.ResourceVersion {
+		return nil
+	}
+	c.addPod(newPod)
+	return nil
+}
+
+func (c *routerController) onDelete(obj interface{}) error {
+	if pod, ok := obj.(*apiv1.Pod); ok {
+		c.addPod(pod)
```
ggarnier: Why do you call `addPod` on `onDelete` method?

provision/kubernetes/cluster.go
cezarsa: Oops, no point, it was from an earlier attempt.

provision/kubernetes/monitor.go
cezarsa: Calling `addPod` means that we're adding the pod app to a queue responsible for ensuring the router targets are correct. This must be called onDelete because the router may still be pointing to a target related to the pod being deleted and if so calling `addPod` will remove such target.

ggarnier:
Closes #2230

PR #2255, opened 2019-01-18
provision/kubernetes: allow dinamically created clusters providers

cezarsa:
Now it's possible to dynamically create kubernetes clusters by leveraging the awesome kontainer-engine project (https://github.com/rancher/kontainer-engine).

Creating new clusters can be done by running:
```
tsuru cluster-add mycluster kubernetes --create-data driver=<driver name> --create-data <driver option>=<option value>
```

It's possible to list available drivers and their options by running`tsuru provisioner-info kubernetes`.


provision/cluster/cluster.go
```diff
@@ -23,49 +24,69 @@ type ClusterProvisioner interface {
 	ClusterHelp() provTypes.ClusterHelpInfo
 }
 
+type CreateClusterProvisioner interface {
```
andrestc: I think `ClusterProvisioner` is more clear (since we can update and delete cluster, `CreateClusterProvisioner` seems a bit off).

cezarsa: There's already a `ClusterProvisioner` so I changed it to `ProvisionClusterProvisioner`

PR #2264, opened 2019-02-05
provision/kubernetes: move deployment and daemonset to v1 from v1beta1

cezarsa:
They are generaly available in apps/v1 since kubernetes 1.9


andrestc:
@cezarsa maybe we should add a compatibility matrix between tsuru and kubernetes somewhere on our docs?

PR #2266, opened 2019-02-07
Save Kubernetes config in app custom resource

ggarnier:
Custom Kubernetes configs from `tsuru.yaml` are now saved in the app CR. The idea is making these configs available to kubernetes-router.

I also extracted `provision.TsuruYamlData` and `router.HealthcheckData` structs to `type/*` packages, so we don't have to add a lot of dependencies when importing those in kubernetes-router project.


ggarnier:
Refs #2213

PR #2283, opened 2019-02-26
provision/kubernetes: refactor informers per cluster

cezarsa:
With @guilhermebr 

This PR simplifies handling informers per cluster and also fixes informers becoming invalid when clusters are updated.

### DIFF
--- a/provision/kubernetes/monitor.go
+++ b/provision/kubernetes/monitor.go
@@ -6,40 +6,50 @@ package kubernetes
 
 import (
 	"sync"
+	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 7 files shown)


provision/kubernetes/monitor.go
```diff
@@ -123,3 +140,79 @@ func (c *routerController) addPod(pod *apiv1.Pod) {
 		rebuild.EnqueueRoutesRebuild(appName)
 	}
 }
+
+func (p *clusterController) getPodInformer() (v1informers.PodInformer, error) {
```
guilhermebr: minor thing, but maybe replace p to c here and in the cases below, just to keep the pattern

PR #2287, opened 2019-03-15
provision/cluster: remove cluster from storage when CreateCluster fails

guilhermebr:


provision/cluster/cluster.go
```diff
@@ -197,6 +197,12 @@ func (s *clusterService) initCluster(c provTypes.Cluster, isNewCluster bool) err
 			err = createProv.UpdateCluster(context.Background(), &c)
 		}
 		if err != nil {
+			if isNewCluster {
+				derr := s.storage.Delete(c)
+				if derr != nil {
+					err = errors.WithStack(derr)
```
cezarsa: We're losing the original `err` here. Maybe something like:
```
err = errors.Wrapf(derr, "error provisioning cluster: %v - error deleting cluster", err)
```

PR #2294, opened 2019-04-16
builder/docker: lookup tsuru.yaml file from same paths as Procfile

### LINKED ISSUE #2133: tsuru.yaml location for image deploys
Currently tsuru searchs for a `tsuru.yaml` file in `/home/application/current` (and some variations) in image deploys. However, we search for `Procfile` in many other additional locations, including `/`.

We should standardize a set of locations to look for both `Procfile` and `tsuru.yaml` files inside an image.

cezarsa:
Fixes #2133

### DIFF
--- a/builder/docker/builder.go
+++ b/builder/docker/builder.go
@@ -343,7 +343,7 @@ func downloadFromURL(url string) (io.ReadCloser, error) {
 	return ioutil.NopCloser(&out), nil
 }
 
-func generateCatCommand(names []string, dirs []string) string {
+func generateCatCommand(names, dirs []string) string {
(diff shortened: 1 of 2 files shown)


builder/docker/builder.go
```diff
@@ -333,3 +342,15 @@ func downloadFromURL(url string) (io.ReadCloser, error) {
 	}
 	return ioutil.NopCloser(&out), nil
 }
+
+func generateCatCommand(names []string, dirs []string) string {
```
nettoclaudio:
```suggestion
func generateCatCommand(names, dirs []string) string {
```

Issue #2295, opened 2019-04-18
Proposal: show related kubernetes-services in app-info


wpjunior:
The internal kubernetes addresses are so hidden by a new user, the user needs to read the documentation and discover the addresses.

I think that the user can figure out theses address via app-info command.

For example:
$ tsuru app-info -a myapp

```
Application: myblog
Platform: gunicorn
Repository: git@githost.com:myblog.git
Teams: team1, team2
Units:
+----------+---------+
| Unit     | State   |
+----------+---------+
| myblog/0 | started |
| myblog/1 | started |


Kubernetes internal addresses:
+----------------------------------------+----------+-------+
| domain                                 | protocol | port  |
+----------------------------------------+----------+-------+
| myblog-web.tsuru-my-pool.cluster.local | tcp      | 8888  |
| myblog-web.tsuru-my-pool.cluster.local | udp      | 12201 |
```

cezarsa:
I agree with this proposal, also I think it would be safe and not too complicated to add this new information to the return of the `GET /apps/:app` API request.

PR #2296, opened 2019-04-20
Show related kubernetes-services in app-info

wpjunior:
Implementation of issue: #2295 

- [x] Initial draft
- [x] Discover app processes
- [x] Tests and mocks
- [x] integrate new provisioner method in HTTP API `GET /apps/:appName`
- [ ] create merge-request to `tsuru-client` using new API.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -5,6 +5,7 @@
 package app
 
 import (
+	"context"
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/json"
@@ -112,6 +113,7 @@ type App struct {
 	UUID string
 
 	// InterApp Properties implemented by provision.InterAppProvisioner
+	// it is lazy generated on the first call to FillInternalAddresses
 	InternalAddresses []*provision.AppInternalAddress `json:",omitempty" bson:"-"`
 
 	Quota       quota.Quota
@@ -134,6 +136,22 @@ func (app *App) getBuilder() (builder.Builder, error) {
 	}
 	app.builder, err = builder.GetForProvisioner(p)
 	return app.builder, err
+}
+
+func (app *App) FillInternalAddresses(ctx context.Context) error {
+	provisioner, err := app.getProvisioner()
+	if err != nil {
+		return err
+	}
+
+	if interAppProvisioner, ok := provisioner.(provision.InterAppProvisioner); ok {
+		app.InternalAddresses, err = interAppProvisioner.InternalAddresses(ctx, app)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
 }
 
 func (app *App) CleanImage(img string) error {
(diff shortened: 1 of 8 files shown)


provision/kubernetes/deploy.go
```diff
@@ -928,6 +928,43 @@ func monitorDeployment(ctx context.Context, client *ClusterClient, dep *appsv1.D
...
+}
+
+func (m *serviceManager) KubernetesInternalAddresses(ctx context.Context, a provision.App) ([]*KubernetesInternalAddress, error) {
+	ns, err := m.client.AppNamespace(a)
+	if err != nil {
+		return nil, err
+	}
+	addresses := []*KubernetesInternalAddress{}
+
+	// TODO: discover app processes
```
andrestc: You can use the App custom resource definition to get this

wpjunior: Thank you so very much! @andrestc

wpjunior:
Hi @andrestc, @guilhermebr, @morpheu  and @cezarsa, could you review my latest changes?

app/app.go
```diff
@@ -111,6 +112,10 @@ type App struct {
 	// UUID is a v4 UUID lazily generated on the first call to GetUUID()
 	UUID string
 
+	// InterApp Properties implemented by provision.InterAppProvisioner
+	// it is lazy generated on the first call to FillInternalAddresses
+	InternalAddresses []*provision.AppInternalAddress `json:",omitempty" bson:"-"`
```
cezarsa: This could be a `[]provision.AppInternalAddress` I don't think there's a need for pointers here.

provision/kubernetes/provisioner.go
```diff
@@ -746,6 +748,53 @@ func (p *kubernetesProvisioner) ListNodes(addressFilter []string) ([]provision.N
...
+
+	processes := []string{}
+	for process := range app.Spec.Services {
+		processes = append(processes, process)
+	}
+	sort.Strings(processes)
+
+	for _, process := range processes {
+		depName := deploymentNameForApp(a, process)
+		service, err := client.CoreV1().Services(ns).Get(depName, metav1.GetOptions{})
```
cezarsa: Would you mind changing this to use the services informer?
It would look something like:
```
controller, err := getClusterController(p, client)
svcInformer, err := controller.getServiceInformer()
for ... {
	service, err := svcInformer.Lister().Services(ns).Get(depName)
```
We didn't update everywhere yet but for new code it's better to stick to informers since they won't cause another hit on the Kubernetes API and will instead read the service from a local cache that's being constantly updated in the background.

wpjunior: Thanks so very much @cezarsa, I've applied all your suggestions. =)

PR #2313, opened 2019-06-04
api: add cluster info route and handler implementation

sacardoso:


api/cluster.go
```diff
@@ -189,6 +189,36 @@ func listClusters(w http.ResponseWriter, r *http.Request, t auth.Token) (err err
...
+//   200: Ok
+//   401: Unauthorized
+//   404: Cluster not found
+func clusterInfo(w http.ResponseWriter, r *http.Request, t auth.Token) (err error) {
+	allowed := permission.Check(t, permission.PermClusterRead)
+	if !allowed {
+		return permission.ErrUnauthorized
+	}
+	var provCluster provTypes.Cluster
+	provCluster.Name = r.URL.Query().Get(":name")
```
guilhermebr: You can remove provCluster and save the name inside a string variable.

```
name := r.URL.Query().Get(":name")
```

sacardoso: Thanks, Guilherme! I've already done the requested change

Issue #2239, opened 2018-12-21
View user quota permission


kasra73:
A non admin user cannot view his quota. As we can see in the code, getUserQuota check for the PermUserUpdateQuota permission. I think there should be be a permission like PermUserViewQuota for this purpose.
![image](https://user-images.githubusercontent.com/7995238/50356468-4510a580-0567-11e9-98b9-d3423df3c17a.png)
https://github.com/tsuru/tsuru/blob/b017fbfddcdb5252d513ad09b1349f88423a8df2/api/quota.go#L31

A normal users has the permission team-member with the roles: context: team, permissions: app.build, app.create, app.delete, app.deploy,app.read, app.run,app.update, service-instance,team,volume.read, volume.update.bind, volume.update.unbind. But a user can only view app quota and not user quota.
I think, as a normal user, I shall be able to view the number of app instances I can create.
But should not be able to update the quota!

PR #2326, opened 2019-07-25
read user quota permission + fix security problem

kasra73:
When I was working on issue #2239 , I found a security problem: Any user could update his/her own quota and no permission is required. This should be true about reading my own quota but not about changing it. So I did these steps:

* Added a user.read.quota permission
* Moved user/email context check from changeUserQuota to getUserQuota function
* Updated the tests

### DIFF
--- a/api/quota.go
+++ b/api/quota.go
@@ -28,11 +28,11 @@ import (
 //   401: Unauthorized
 //   404: User not found
 func getUserQuota(w http.ResponseWriter, r *http.Request, t auth.Token) error {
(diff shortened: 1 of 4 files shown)


cezarsa:
Thanks a lot for your contribution and for catching this. The root problem is that we add an implicit permission for all users on `user.*` for their own user. This makes sense almost everywhere, except for `user.update.quota`. I think that in the future we should move the quota permission to another permission tree to avoid repeating this kind of bug in other places.

Issue #2257, opened 2019-01-30
Show server version in `tsuru version` command


ggarnier:
Currently `tsuru version` command show only the client version. It could fetch `/info` endpoint and display the server version as well.

nettoclaudio:
Closed due #2336 and https://github.com/tsuru/tsuru-client/pull/143.

PR #2336, opened 2019-09-10
Returns server version on: /info route

arthurcgc:
Closes Issue #2257
tsuru --version now shows Server version as well as Client version. If it can't establish a connection to the server, it raises an error showing that the server is probably down.

### DIFF
--- a/api/info.go
+++ b/api/info.go
@@ -7,6 +7,8 @@ package api
 import (
 	"encoding/json"
 	"net/http"
+
+	"github.com/tsuru/tsuru/auth"
 )
 
 // title: api info
@@ -15,7 +17,7 @@ import (
 // produce: application/json
 // responses:
 //   200: OK
-func info(w http.ResponseWriter, r *http.Request) error {
+func info(w http.ResponseWriter, r *http.Request, t auth.Token) error {
 	data := map[string]string{}
 	data["version"] = Version
 	w.Header().Set("Content-Type", "application/json")
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -622,37 +622,40 @@ func versionString(manager *Manager) string {
 
 func apiVersionString(client *Client) (string, error) {
 	if client == nil {
-		return "Server not found", nil
+		return "", fmt.Errorf("Null Client Exception")
 	}
+
 	url, err := GetURL("/info")
 	if err != nil {
 		return "", err
 	}
+
 	req, err := http.NewRequest("GET", url, nil)
 	if err != nil {
 		return "", err
 	}
+
+	serverVersionPrefix := "Server version: "
 	resp, err := client.Do(req)
 	if err != nil {
+		if isUnauthorized(err) {
+			return fmt.Sprintf("%s Can't retrieve server version, user %v", serverVersionPrefix, err), nil
+		}
 		return "", err
 	}
 
 	body, err := ioutil.ReadAll(resp.Body)
-	if err != nil {
-		return "", err
-	}
-	err = resp.Body.Close()
+	defer resp.Body.Close()
 	if err != nil {
 		return "", err
 	}
 
 	var version map[string]string
-
 	err = json.Unmarshal(body, &version)
 	if err != nil {
 		return "", err
 	}
-	return fmt.Sprintf("Server version: %s\n", version["version"]), nil
+	return fmt.Sprintf("%s%s\n", serverVersionPrefix, version["version"]), nil
 }
 
 func (c *version) Run(context *Context, client *Client) error {
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -622,7 +622,7 @@ func versionString(manager *Manager) string {
 
 func apiVersionString(client *Client) (string, error) {
 	if client == nil {
-		return "", fmt.Errorf("Null Client Exception")
+		return "", fmt.Errorf("client cannot be nil")
 	}
 
 	url, err := GetURL("/info")
@@ -635,13 +635,10 @@ func apiVersionString(client *Client) (string, error) {
 		return "", err
 	}
 
-	serverVersionPrefix := "Server version: "
 	resp, err := client.Do(req)
 	if err != nil {
-		if isUnauthorized(err) {
-			return fmt.Sprintf("%s Can't retrieve server version, user %v", serverVersionPrefix, err), nil
-		}
-		return "", err
+		// if we return the error, stdout won't flush until the prompt
+		return fmt.Sprintf("Unable to retrieve server version: %v", err), nil
 	}
 
 	body, err := ioutil.ReadAll(resp.Body)
@@ -655,7 +652,7 @@ func apiVersionString(client *Client) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	return fmt.Sprintf("%s%s\n", serverVersionPrefix, version["version"]), nil
+	return fmt.Sprintf("Server version: %s\n", version["version"]), nil
 }
 
 func (c *version) Run(context *Context, client *Client) error {
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -642,7 +642,6 @@ func apiVersionString(client *Client) (string, error) {
 	}
 
 	body, err := ioutil.ReadAll(resp.Body)
-	defer resp.Body.Close()
 	if err != nil {
 		return "", err
 	}
@@ -652,7 +651,9 @@ func apiVersionString(client *Client) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	return fmt.Sprintf("Server version: %s\n", version["version"]), nil
+
+	resp.Body.Close()
+	return fmt.Sprintf("Server version: %s.\n", version["version"]), nil
 }
 
 func (c *version) Run(context *Context, client *Client) error {
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -621,39 +621,38 @@ func versionString(manager *Manager) string {
 }
 
 func apiVersionString(client *Client) (string, error) {
+	if client == nil {
+		return "API Server not found", nil
+	}
 	url, err := GetURL("/info")
 	if err != nil {
 		return "", err
 	}
 	req, err := http.NewRequest("GET", url, nil)
 	if err != nil {
-		return "", nil
+		return "", err
 	}
 	resp, err := client.Do(req)
-	if err != nil {
-		return "", nil
-	}
-	// if resp.StatusCode == http.StatusOK {
-	// 	return fmt.Sprintf("API version: %s\n", resp.Body), nil
-	// }
-
-	body, err := ioutil.ReadAll(resp.Body)
-	resp.Body.Close()
-	if err != nil {
-		return "", nil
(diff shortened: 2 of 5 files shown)


magnotorres:
Only make sure if you will display the server version to authenticated users, for security reasons

cmd/cmd.go
```diff
@@ -614,11 +617,56 @@ func versionString(manager *Manager) string {
...
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return "", err
+	}
+
+	serverVersionPrefix := "Server version: "
+	resp, err := client.Do(req)
+	if err != nil {
+		if isUnauthorized(err) {
+			return fmt.Sprintf("%s Can't retrieve server version, user %v", serverVersionPrefix, err), nil
```
nettoclaudio: Instead handling only the unauthorized case, what about turning this error message more generic to fit w/ any error? I'm thinking something like: `fmt.Sprintf("Unable to retrieve server version: %v", err)`.

I also think it's better suppress the `serverVersionPrefix` on error cases... I had suggested just the opposite before, my bad.

cmd/cmd.go
```diff
@@ -614,11 +617,56 @@ func versionString(manager *Manager) string {
 	if GitHash != "" {
 		suffix = fmt.Sprintf(" hash %s\n", GitHash)
 	}
-	return fmt.Sprintf("%s version %s.%s", manager.name, manager.version, suffix)
+	return fmt.Sprintf("Client version: %s.%s", manager.version, suffix)
+}
+
+func apiVersionString(client *Client) (string, error) {
+	if client == nil {
+		return "", fmt.Errorf("Null Client Exception")
```
nettoclaudio: This error message looks like a Java exception :nauseated_face: I would used something like `client cannot be nil`.

cmd/cmd_test.go
```diff
@@ -1189,3 +1190,66 @@ func (c *FailCommandCustom) Info() *Info {
...
+	context := Context{[]string{}, mngr.stdout, mngr.stderr, mngr.stdin}
+
+	ts := httptest.NewServer(http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
+		// Send response to be tested
+		rw.Write([]byte(`{"version":"1.7.4"}`))
+	}))
+
+	client := NewClient(ts.Client(), &context, mngr)
+	err := command.Run(&context, client)
+	c.Assert(err, check.NotNil)
```
nettoclaudio: I think you forget the assertions about the client/server versions here... :smile:

cmd/cmd_test.go
```diff
@@ -1189,3 +1190,66 @@ func (c *FailCommandCustom) Info() *Info {
...
+	context := Context{[]string{}, mngr.stdout, mngr.stderr, mngr.stdin}
+
+	ts := httptest.NewServer(http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
+		// Send response to be tested
+		rw.Write([]byte(`{"version":"1.7.4"}`))
+	}))
+
+	client := NewClient(&http.Client{}, &context, mngr)
+
+	ts.URL = "faketsuru.io"
```
nettoclaudio: You should use another TLD like `.test` as recommended by RFC 2606 (https://tools.ietf.org/html/rfc2606.html#section-2)

cmd/cmd_test.go
```diff
@@ -1189,3 +1190,66 @@ func (c *FailCommandCustom) Info() *Info {
...
+	os.Setenv("TSURU_TARGET", ts.URL)
+	defer os.Unsetenv("TSURU_TARGET")
+	err := command.Run(&context, client)
+	c.Assert(err, check.NotNil)
```
nettoclaudio: You could check if the client version is still shown and also the the expected message returned when the API is unreachable.

cmd/cmd_test.go
```diff
@@ -1189,3 +1190,66 @@ func (c *FailCommandCustom) Info() *Info {
...
+
+func (s *S) TestVersionAPIInvalidURL(c *check.C) {
+	var stdout, stderr bytes.Buffer
+	mngr := NewManager("tsuru", "5.0", "", &stdout, &stderr, os.Stdin, nil)
+	var exiter recordingExiter
+	mngr.e = &exiter
+	command := version{manager: mngr}
+	context := Context{[]string{}, mngr.stdout, mngr.stderr, mngr.stdin}
+
+	ts := httptest.NewServer(http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
```
nettoclaudio: You don't have to instantiate a testing server for this scenario.

cmd/cmd_test.go
```diff
@@ -1189,3 +1190,66 @@ func (c *FailCommandCustom) Info() *Info {
...
+	client := NewClient(ts.Client(), &context, mngr)
+	err := command.Run(&context, client)
+	c.Assert(err, check.NotNil)
+}
+
+func (s *S) TestVersionAPIInvalidURL(c *check.C) {
+	var stdout, stderr bytes.Buffer
+	mngr := NewManager("tsuru", "5.0", "", &stdout, &stderr, os.Stdin, nil)
+	var exiter recordingExiter
+	mngr.e = &exiter
```
nettoclaudio: You may checking whether exit code is zero.

cmd/cmd_test.go
```diff
@@ -1189,3 +1190,66 @@ func (c *FailCommandCustom) Info() *Info {
...
+	command := version{manager: mngr}
+	context := Context{[]string{}, mngr.stdout, mngr.stderr, mngr.stdin}
+
+	ts := httptest.NewServer(http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
+		// Send response to be tested
+		rw.Write([]byte(`{"version":"1.7.4"}`))
+	}))
```
nettoclaudio: You forget to close the testing server.

cmd/cmd_test.go
```diff
@@ -1189,3 +1190,66 @@ func (c *FailCommandCustom) Info() *Info {
...
```
nettoclaudio: Same as above, close this guy after the test execution.

cmd/cmd_test.go
```diff
@@ -1189,3 +1190,66 @@ func (c *FailCommandCustom) Info() *Info {
...
+func (s *S) TestVersionWithAPI(c *check.C) {
+	var stdout, stderr bytes.Buffer
+	mngr := NewManager("tsuru", "5.0", "", &stdout, &stderr, os.Stdin, nil)
+	var exiter recordingExiter
+	mngr.e = &exiter
+	command := version{manager: mngr}
+	context := Context{[]string{}, mngr.stdout, mngr.stderr, mngr.stdin}
+
+	ts := httptest.NewServer(http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
+		// Send response to be tested
```
nettoclaudio: Drop this comment :)

cmd/cmd_test.go
```diff
@@ -1189,3 +1190,66 @@ func (c *FailCommandCustom) Info() *Info {
...
+func (s *S) TestVersionWithoutEnvVar(c *check.C) {
+	var stdout, stderr bytes.Buffer
+	mngr := NewManager("tsuru", "5.0", "", &stdout, &stderr, os.Stdin, nil)
+	var exiter recordingExiter
+	mngr.e = &exiter
+	command := version{manager: mngr}
+	context := Context{[]string{}, mngr.stdout, mngr.stderr, mngr.stdin}
+
+	ts := httptest.NewServer(http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
+		// Send response to be tested
```
nettoclaudio: Drop this comment :)

cmd/cmd.go
```diff
@@ -614,11 +617,53 @@ func versionString(manager *Manager) string {
...
+	}
+
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return "", err
+	}
+
+	resp, err := client.Do(req)
+	if err != nil {
+		// if we return the error, stdout won't flush until the prompt
```
nettoclaudio: Nice explanation. :+1:

cmd/cmd.go
```diff
@@ -614,11 +617,53 @@ func versionString(manager *Manager) string {
...
+		return fmt.Sprintf("Unable to retrieve server version: %v", err), nil
+	}
+
+	body, err := ioutil.ReadAll(resp.Body)
+	defer resp.Body.Close()
```
nettoclaudio: Only makes sense close the Body reader when no error is returned, move this statement to below of error checking.

cmd/cmd_test.go
```diff
@@ -1189,3 +1190,68 @@ func (c *FailCommandCustom) Info() *Info {
...
+	c.Assert(mngr.stdout.(*bytes.Buffer).String(),
+		check.Equals, "Client version: 5.0.\nUnable to retrieve server version: Failed to connect to tsuru server (http://localhost), it's probably down.")
+}
+
+func (s *S) TestVersionAPIInvalidURL(c *check.C) {
+	var stdout, stderr bytes.Buffer
+	mngr := NewManager("tsuru", "5.0", "", &stdout, &stderr, os.Stdin, nil)
+	var exiter recordingExiter
+	mngr.e = &exiter
+	c.Assert(exiter.value(), check.Equals, 0)
```
nettoclaudio: This assertion should be made after the command execution (at line 1253)

cmd/cmd.go
```diff
@@ -614,11 +617,53 @@ func versionString(manager *Manager) string {
...
+	if err != nil {
+		return "", err
+	}
+
+	var version map[string]string
+	err = json.Unmarshal(body, &version)
+	if err != nil {
+		return "", err
+	}
+	return fmt.Sprintf("Server version: %s\n", version["version"]), nil
```
nettoclaudio: You could add a dot at end of this sentence, just for looks like the client sentence.

cmd/cmd_test.go
```diff
@@ -1189,3 +1190,66 @@ func (c *FailCommandCustom) Info() *Info {
...
+	defer ts.Close()
+
+	client := NewClient(&http.Client{}, &context, mngr)
+
+	os.Setenv("TSURU_TARGET", ts.URL)
+	defer os.Unsetenv("TSURU_TARGET")
+	err := command.Run(&context, client)
+	c.Assert(err, check.IsNil)
+	c.Assert(mngr.stdout.(*bytes.Buffer).String(),
+		check.Equals, "Client version: 5.0.\nServer version: 1.7.4\n")
```
nettoclaudio:
```suggestion
		check.Equals, "Client version: 5.0.\nServer version: 1.7.4.\n")
```

api/info.go
```diff
@@ -15,7 +17,7 @@ import (
 // produce: application/json
 // responses:
 //   200: OK
-func info(w http.ResponseWriter, r *http.Request) error {
+func info(w http.ResponseWriter, r *http.Request, t auth.Token) error {
```
nettoclaudio: Now the CI is failing because this route is authenticated and the test `TestInfo` isn't prepared for.

Issue #1914, opened 2018-01-17
Service instance ops (bind/unbind/...) should include TeamOwner in permissions


andrestc:
Whenever checking permissions for service instance ops, we validate if the user has `PermServiceInstanceUpdateUnbind` on `instance.Teams`. We should include the `instace.TeamOwner` in this check.

JonathanAlcantara:
I'm giving a look at this issue.

andrestc:
Thanks, @JonathanAlcantara. Reach out if you need any help.

dnguy078:
@andrestc could you review this https://github.com/tsuru/tsuru/pull/2350

PR #2350, opened 2019-10-01
Permissions check

dnguy078:
- Fixes issue: https://github.com/tsuru/tsuru/issues/1914


dnguy078:
@nettoclaudio I'll update that too, thank you!

PR #2352, opened 2019-10-03
Add Log Level

juliolustosa:
Filter logs by level

### DIFF
--- a/applog/memory.go
+++ b/applog/memory.go
@@ -118,7 +118,8 @@ func (s *memoryLogService) List(args appTypes.ListLogArgs) ([]appTypes.Applog, e
 	var count int
 	for current := buffer.end; count < args.Limit; {
 		if (args.Source == "" || (args.Source == current.log.Source) != args.InvertFilters) &&
-			(args.Unit == "" || (args.Unit == current.log.Unit) != args.InvertFilters) {
+			(args.Unit == "" || (args.Unit == current.log.Unit) != args.InvertFilters) &&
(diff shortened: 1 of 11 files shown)


types/app/log.go
```diff
@@ -14,7 +14,7 @@ type LogWatcher interface {
 
 type AppLogService interface {
 	Enqueue(entry *Applog) error
-	Add(appName, message, source, unit string) error
+	Add(appName, message, source, unit string, level int) error
```
cezarsa: I don't think it's a good idea to change this method signature. This method is only a helper used for log messages produced by tsuru itself. We can assume they will always have the same level and leave this method as it was before, reducing the diff for this PR.

applog/memory.go
```diff
@@ -118,7 +119,8 @@ func (s *memoryLogService) List(args appTypes.ListLogArgs) ([]appTypes.Applog, e
 	var count int
 	for current := buffer.end; count < args.Limit; {
 		if (args.Source == "" || (args.Source == current.log.Source) != args.InvertFilters) &&
-			(args.Unit == "" || (args.Unit == current.log.Unit) != args.InvertFilters) {
+			(args.Unit == "" || (args.Unit == current.log.Unit) != args.InvertFilters) &&
+			(args.Level == 0 || (args.Level == current.log.Level) != args.InvertFilters) {
```
cezarsa: We need some tests for this check and also for the equivalent change on the mongodb implementation. Adding a new method to the `ServiceSuite` in `applog/service_test.go` will automatically run the test on both implementations.

Issue #2364, opened 2019-10-17
make test failed


nanikjava:
**Description**
> make test is failing 

**Steps to reproduce**
> make test

**Expected result**
> no test should failed

**Attachments**
> If applicable, provide logs, screenshots, screencast, etc to help explain your problem.

nanikjava:
The test is failing inside TestDeleteMachineErrorNoServer method 

Will submit PR to fix this

nanikjava:
How come running in travis yields different result

```
FAIL: iaas_test.go:358: cloudstackSuite.TestDeleteMachineErrorNoServer

iaas_test.go:365:

    c.Assert(err, check.ErrorMatches, ".*Temporary failure in name resolution*")

... error string = "Get http://invalidurl.invalid.invalid?apiKey=test&command=listVolumes&response=json&virtualmachineid=myMachineId&signature=A4dPF7Y85uOmkoqbnTK8fa9OUug%3D: dial tcp: lookup invalidurl.invalid.invalid on 169.254.169.254:53: no such host"

... regex string = ".*Temporary failure in name resolution*"

OOPS: 13 passed, 1 FAILED

--- FAIL: Test (14.15s)
```

PR #2365, opened 2019-10-17
TestDeleteMachineErrorNoServer is failing

nanikjava:
Fixes : #2364 

Running make test failed inside TestDeleteMachineErrorNoServer method.

Looks like the returned error message has been modified


cezarsa:
This change breaks the tests on CI ideally it should work anywhere.

PR #2382, opened 2020-01-03
Ensure image versions are never reused and always incremented

cezarsa:
It was possible for a failed build to reuse the same version multiple
times for each failed attempt. This happened because the version counter
was only incremeneted *after* the build step and just before the deploy
step.

The changes introduced here ensure we always increment the version
counter in the database before any other action is taken, this
incremented version is then propagated to the provisioner.

### DIFF
--- a/api/build_test.go
+++ b/api/build_test.go
@@ -142,10 +142,10 @@ func (s *BuildSuite) SetUpTest(c *check.C) {
 }
 
 func (s *BuildSuite) TestBuildHandler(c *check.C) {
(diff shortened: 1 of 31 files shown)


net/client.go
```diff
@@ -24,6 +24,9 @@ func makeTimeoutHTTPClient(dialTimeout time.Duration, fullTimeout time.Duration,
 			TLSHandshakeTimeout: dialTimeout,
 			MaxIdleConnsPerHost: maxIdle,
 			IdleConnTimeout:     15 * time.Second,
+			TLSClientConfig: &tls.Config{
+				InsecureSkipVerify: true,
+			},
```
nettoclaudio: Should this change really be tracked?  :see_no_evil:

cezarsa: oops

PR #2389, opened 2020-01-29
Event webhook exec register

### LINKED ISSUE #2098: Register event webhook executions
It would be nice if we registered events upon event webhook executions (something like `webhook.run`). The problem with this approach is that the number of events could grow exponentially.

Also, we'd need to block `webhook.run` events triggering other webhooks!

arthurcgc:
This PR adds a feature that enables tsuru to Log in `tsuru event-list` an event when a Webhook is triggered, it also makes sure that the user is unable to create a Webhook that triggers another Webhook, that would create an infinite loop that would flood logged events in tsuru. 

Closes #2098.

### DIFF
--- a/api/webhook.go
+++ b/api/webhook.go
@@ -17,6 +17,8 @@ import (
 	permTypes "github.com/tsuru/tsuru/types/permission"
 )
 
+var errBadWebhookRequest = &tsuruErrors.HTTP{Code: http.StatusBadRequest, Message: "can't create a webhook that triggers another webhook"}
+
 // title: webhook list
 // path: /events/webhooks
 // method: GET
@@ -93,7 +95,7 @@ func webhookCreate(w http.ResponseWriter, r *http.Request, t auth.Token) error {
 	}
 	for _, kindName := range webhook.EventFilter.KindNames {
 		if kindName == "webhook.run" {
-			return &tsuruErrors.HTTP{Code: http.StatusBadRequest, Message: "can't create a webhook that triggers another webhook"}
+			return errBadWebhookRequest
 		}
 	}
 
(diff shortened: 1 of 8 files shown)


api/webhook.go
```diff
@@ -90,6 +90,12 @@ func webhookCreate(w http.ResponseWriter, r *http.Request, t auth.Token) error {
 			return err
 		}
 	}
+	for _, kindName := range webhook.EventFilter.KindNames {
+		if kindName == "webhook.run" {
+			return permission.ErrUnauthorized
```
bernardolins: I don't think unauthorized is the best error type for this case. The error is caused not because of a lack of permissions, but because this kind of webhook is not allowed. What do you think of a bad request?

types/event/webhook.go
```diff
@@ -12,6 +12,7 @@ import (
 var (
 	ErrWebhookAlreadyExists = errors.New("webhook already exists with the same name")
 	ErrWebhookNotFound      = errors.New("webhook not found")
+	ErrWebhookLoop          = errors.New("webhook can't trigger another webhook")
```
bernardolins: You are not using this anywhere, I think it's safe to remove.

api/webhook.go
```diff
@@ -133,6 +140,13 @@ func webhookUpdate(w http.ResponseWriter, r *http.Request, t auth.Token) error {
 	if !permission.Check(t, permission.PermWebhookUpdate, ctx) {
 		return permission.ErrUnauthorized
 	}
+
+	for _, kindName := range webhook.EventFilter.KindNames {
+		if kindName == "webhook.run" {
+			return permission.ErrUnauthorized
```
bernardolins: Like in `webhookCreate`, I think unauthorized is not an appropriate response here.

PR #2400, opened 2020-05-15
Improve container image garbage collector: add metrics, locks and creates a tsuru event

wpjunior:


app/image/gc/gc.go
```diff
@@ -10,20 +10,78 @@ import (
...
+var (
+	gcExecutionsTotal = promauto.NewCounterVec(prometheus.CounterOpts{
+		Name: "tsuru_gc_executions_total",
+		Help: "The number of times that gc had runned by phase",
+	}, []string{"phase"})
+
+	executionDuration = promauto.NewHistogramVec(prometheus.HistogramOpts{
+		Name:    "tsuru_gc_execution_duration_seconds",
+		Help:    "How long during the GC process",
+		Buckets: prometheus.ExponentialBuckets(0.1, 3, 10),
```
cezarsa, line 43: nitpick: Upper bound is 0.1*(3^10) ~= 1.6 hours. Maybe a bit excessive. :)  What about a 2.7 factor for ~30 minutes upper bound and better resolution or less buckets?

app/image/gc/gc.go
```diff
@@ -67,111 +123,331 @@ func (g *imgGC) spin() {
...
 		}
 		if a == nil {
-			log.Debugf("[image gc] app %q not found, removing everything", versions.AppName)
-			err = registry.RemoveAppImages(versions.AppName)
-			if err != nil {
-				multi.Add(err)
-			}
-			err = servicemanager.AppVersion.DeleteVersions(versions.AppName)
+			log.Debugf("[image gc] app %q not found, removing everything", appVersions.AppName)
+			err = pruneAllVersionsByApp(appVersions.AppName)
```
cezarsa, line 211: Since we're separating mark and sweep this seems a bit out of place here. Wouldn't it be better to only mark all these versions instead of sweeping straight away?

app/image/gc/gc.go
```diff
@@ -67,111 +123,331 @@ func (g *imgGC) spin() {
...
+	versionsToRemove, versionsToPruneFromProvisioner := selectAppVersions(appVersions, deployedVersions, historySize)
+	if !exclusiveLockAcquired {
+		for _, version := range versionsToPruneFromProvisioner {
+			pruneVersionFromProvisioner(a, version)
+		}
+	}
+
+	if len(versionsToRemove) == 0 {
+		return false, nil
+	} else if !exclusiveLockAcquired {
```
cezarsa, line 268: nitpick: No need for `else` here. When reading the code the `else` statement implies there's a relation between this block and the previous if statement when actually they are not related, the `if len(versionsToRemove) == 0` is a short-circuit and unrelated to the next short circuit condition that is the lock. IMO it's easier to read when they are separated because I don't have to pause and think about how the two conditions are related:
```go
if len(versionsToRemove) == 0 {
	return false, nil
}

if !exclusiveLockAcquired {
	return true, nil
}
```

app/image/gc/gc.go
```diff
@@ -67,111 +123,331 @@ func (g *imgGC) spin() {
...
+		return true, nil
+	}
+
+	for _, version := range versionsToRemove {
+		versionsMarkedToRemovalTotal.Inc()
+		pruned := pruneVersionFromProvisioner(a, version)
```
cezarsa, line 274: What do you think of moving this to the sweep step? The reasoning would be to keep the lock for as little time as possible and this call could potentially take a while. The previous call for pruneVersionFromProvisioner in this function is fine because we're not locking yet.

PR #2404, opened 2020-06-02
feat: Adding WSL2 oauth support

markiing:
Checking if user is in WSL env and executes powershell to open GUI instead of xdg.

### DIFF
--- a/cmd/open.go
+++ b/cmd/open.go
@@ -6,12 +6,37 @@
 
 package cmd
 
-import "github.com/tsuru/tsuru/exec"
+import (
+	"fmt"
+	"strings"
+	"golang.org/x/sys/unix"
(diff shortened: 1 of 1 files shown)


cmd/open.go
```diff
@@ -6,11 +6,33 @@
...
+	release := strings.ToLower(string(u.Release[:]))
+	return strings.Contains(release, "microsoft")
+}
 
 func open(url string) error {
+
+	cmd := "xdg-open"
+
+	if(isWSL()){
+		cmd = "powershell.exe -c start"
```
wpjunior, line 31: @markiing, I supposed that need to separate cmd in args,
example
```
cmd = "powershell.exe"
args = []string{"-c", "start", url"}
```

markiing: Done :D

PR #2405, opened 2020-06-09
Add a flag to suppress sensitive environment variables at the event API

wpjunior:


app/security.go
```diff
@@ -0,0 +1,50 @@
...
+		}
+	}
+
+	return false
+}
+
+func (a *App) SuppressSensitiveEnvs() {
+	newEnv := map[string]bind.EnvVar{}
+	for key, env := range a.Env {
+		if !env.Public || isSensitiveName(env.Name) {
```
nettoclaudio, line 38: It's better to rely only on the state declared by the user. Deducing whether the env var is sensitive based on its name may cause misunderstanding.

```suggestion
		if !env.Public {
```

app/security.go
```diff
@@ -0,0 +1,50 @@
+package app
```
nettoclaudio, line 1: Remember the license header...

app/security.go
```diff
@@ -0,0 +1,50 @@
...
+	for key, env := range a.Env {
+		if !env.Public || isSensitiveName(env.Name) {
+			env.Value = suppressedEnv
+		}
+		newEnv[key] = env
+	}
+	a.Env = newEnv
+
+	for i, serviceEnv := range a.ServiceEnvs {
+		if !serviceEnv.EnvVar.Public || isSensitiveName(serviceEnv.EnvVar.Name) {
```
nettoclaudio, line 46: IDK whether the services can inject a private env var, I hope so. Above suggestion applies here too.

PR #2408, opened 2020-06-17
Stream logs directly from kubernetes api-server

wpjunior:
This MR is an architectural change using kubernetes provisioner, we intend to maintain Tsuru as a globally control-plane application, we are assigning the responsibility to store logs on kubelet engine, then allowing Tsuru to maintain many k8s clusters without bottleneck.

### DIFF
--- a/provision/kubernetes/logs.go
+++ b/provision/kubernetes/logs.go
@@ -93,6 +93,9 @@ func (p *kubernetesProvisioner) WatchLogs(app appTypes.App, args appTypes.ListLo
 	if err != nil {
 		return nil, err
 	}
+	if len(args.Units) > 0 {
+		pods = filterPods(pods, args.Units)
+	}
 	watchingPods := map[string]bool{}
 	ctx, done := context.WithCancel(context.Background())
 	watcher := &k8sLogsNotifier{
@@ -107,15 +110,17 @@ func (p *kubernetesProvisioner) WatchLogs(app appTypes.App, args appTypes.ListLo
 			continue
 		}
 		watchingPods[pod.ObjectMeta.Name] = true
+		watcher.wg.Add(1)
 		go watcher.watchPod(pod, false)
 	}
 
 	podListener := &podListener{
 		OnEvent: func(pod *apiv1.Pod) {
 			_, alreadyWatching := watchingPods[pod.ObjectMeta.Name]
-
-			if !alreadyWatching && pod.Status.Phase != apiv1.PodPending {
+			podMatches := matchPod(pod, args)
+			if !alreadyWatching && podMatches && pod.Status.Phase != apiv1.PodPending {
 				watchingPods[pod.ObjectMeta.Name] = true
+				watcher.wg.Add(1)
 				go watcher.watchPod(pod, true)
 			}
 		},
@@ -271,7 +276,7 @@ type k8sLogsNotifier struct {
 }
 
 func (k *k8sLogsNotifier) watchPod(pod *apiv1.Pod, notify bool) {
-	k.wg.Add(1)
+
 	defer k.wg.Done()
 	appName := pod.ObjectMeta.Labels["tsuru.io/app-name"]
 	appProcess := pod.ObjectMeta.Labels["tsuru.io/app-process"]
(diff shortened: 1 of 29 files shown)


provision/kubernetes/logs.go
```diff
@@ -0,0 +1,318 @@
...
+	knet "k8s.io/apimachinery/pkg/util/net"
+)
+
+const (
+	logLineTimeSeparator = " "
+	logWatchBufferSize   = 1000
+)
+
+var (
+	watchNewPodsInterval = time.Second * 10
```
cezarsa, line 29: unused

provision/kubernetes/logs.go
```diff
@@ -0,0 +1,318 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	podInformer, err := clusterController.getPodInformer()
+	if err != nil {
+		return nil, err
+	}
+
+	pods, err := podInformer.Lister().Pods(ns).List(listPodsSelectorForLog(args))
```
cezarsa, line 56: Missing `err` handling.

provision/kubernetes/logs.go
```diff
@@ -0,0 +1,318 @@
...
+	ns   string
+	wg   sync.WaitGroup
+	once sync.Once
+	done context.CancelFunc
+
+	clusterClient *ClusterClient
+}
+
+func (k *k8sLogsNotifier) watchPod(pod *apiv1.Pod, notify bool) {
+	k.wg.Add(1)
```
cezarsa, line 264: This is dangerous and could cause unexpected behavior. Waitgroups have one weird quirk, according to the [docs](https://golang.org/pkg/sync/#WaitGroup.Add):

> Note that calls with a positive delta that occur when the counter is zero must happen before a Wait.

Since this is called in a goroutine there's a chance that this code will only be executed after the call to Wait(). This would cause the Wait() call to return immediately and assume everything has finished.

I think it would be safer to increment the waitgroup in the caller, before starting the goroutine.

provision/kubernetes/logs.go
cezarsa: Also, we must filter `pods` by their IDs using `args.Units` here.

provision/kubernetes/logs.go
```diff
@@ -0,0 +1,318 @@
...
+
+		unifiedLog = append(unifiedLog, errToLog(pod.ObjectMeta.Name, appName, err))
+	}
+
+	return unifiedLog, nil
+}
+
+func listPodsSelectorForLog(args appTypes.ListLogArgs) labels.Selector {
+	return labels.SelectorFromSet(labels.Set(map[string]string{
+		"tsuru.io/app-name": args.AppName,
```
cezarsa, line 203: We must use `args.Source` if set to filter by `tsuru.io/app-process` here.

provision/kubernetes/logs.go
```diff
@@ -0,0 +1,318 @@
...
+			_, alreadyWatching := watchingPods[pod.ObjectMeta.Name]
+
+			if !alreadyWatching && pod.Status.Phase != apiv1.PodPending {
+				watchingPods[pod.ObjectMeta.Name] = true
+				go watcher.watchPod(pod, true)
+			}
+		},
+	}
+
+	clusterController.addPodListener(args.AppName, podListener)
```
cezarsa, line 118: Instead of calling `addPodListener` with a `*podListener` what do you think about having it declared as:
```
func (controller) addPodListener(appName string, listener interface{ OnEvent(*apiv1.Pod) })
```
And then implementing the `OnEvent` handler in the `k8sLogsNotifier` itself. This way we would avoid this closure here and simply call:
```
clusterController.addPodListener(args.appName, watcher)
```

provision/kubernetes/monitor.go
```diff
@@ -47,6 +47,11 @@ var eventKindsIgnoreRebuild = []string{
 	permission.PermAppUpdateRoutable.FullName(),
 }
 
+type podListener struct {
+	OnEvent func(*apiv1.Pod)
+}
+type podListeners map[*podListener]struct{}
```
cezarsa, line 53: Unused

PR #2409, opened 2020-06-18
feat(api/service): support updating and showing plan parameters for service instances

nettoclaudio:


wpjunior:
LGTM @nettoclaudio

PR #2410, opened 2020-06-19
fix(api/service): support for force removal of service instances

nettoclaudio:


cezarsa:
Maybe we're mixing two different things here. I think the existing request parameter `unbindall` should only mean that we'll unbind all apps from the instance first. Assuming that we'll also ignore service API errors because `unbindall` is set can be misleading.
What do you think about having a separated request parameter (something like `ignoreerrors`) and using it to control ignoring service API errors?

nettoclaudio:
I absolutely agree w/ you, @cezarsa... I'd considered doing it like you've suggested but the way `tsuru-client` handles this parameter on CLI (as `--force` flag) led me to reuse this `unbindall` query parameter. Should do I also create a new flag on `tsuru-client` to activate the `ignoreerrors`?

cezarsa:
Yeah, I think the client could go with `--force-unbind` and `--force-removal`.

PR #2411, opened 2020-06-30
provision/kubernetes: add option to use entire cluster as a single pool

morpheu:
This PR adds a new `single-pool` option to cluster that allows using the entire cluster as a single Tsuru pool

### DIFF
--- a/provision/kubernetes/nodecontainer.go
+++ b/provision/kubernetes/nodecontainer.go
@@ -93,6 +93,13 @@ func (m *nodeContainerManager) deployNodeContainerForCluster(client *ClusterClie
 			},
 		},
 	}
+	singlePool, err := client.SinglePool(pool)
+	if err != nil {
+		return err
+	}
+	if singlePool && pool != "" {
+		affinity = &apiv1.Affinity{}
+	}
 	if oldDs != nil && placementOnly {
 		if reflect.DeepEqual(oldDs.Spec.Template.Spec.Affinity, affinity) {
 			return nil
--- a/provision/kubernetes/nodecontainer.go
+++ b/provision/kubernetes/nodecontainer.go
@@ -97,7 +97,10 @@ func (m *nodeContainerManager) deployNodeContainerForCluster(client *ClusterClie
 	if err != nil {
 		return err
 	}
-	if singlePool && pool != "" {
+	if singlePool {
+		if pool == "" {
+			return nil
+		}
 		affinity = &apiv1.Affinity{}
 	}
 	if oldDs != nil && placementOnly {
--- a/provision/kubernetes/nodecontainer.go
+++ b/provision/kubernetes/nodecontainer.go
@@ -93,7 +93,7 @@ func (m *nodeContainerManager) deployNodeContainerForCluster(client *ClusterClie
 			},
 		},
 	}
-	singlePool, err := client.SinglePool(pool)
+	singlePool, err := client.SinglePool()
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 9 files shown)


provision/kubernetes/nodecontainer.go
```diff
@@ -93,6 +93,13 @@ func (m *nodeContainerManager) deployNodeContainerForCluster(client *ClusterClie
 			},
 		},
 	}
+	singlePool, err := client.SinglePool()
+	if err != nil {
+		return err
```
wpjunior, line 98: @morpheu could you use errors.WithMessage here ?

provision/kubernetes/nodecontainer.go
```diff
@@ -93,6 +93,13 @@ func (m *nodeContainerManager) deployNodeContainerForCluster(client *ClusterClie
...
+	}
+	if singlePool && pool != "" {
```
wpjunior, line 100: why we need to evaluate `pool != ""` ?

provision/kubernetes/provisioner.go
```diff
@@ -201,6 +201,9 @@ func (p *kubernetesProvisioner) InitializeCluster(c *provTypes.Cluster) error {
 }
 
 func (p *kubernetesProvisioner) ValidateCluster(c *provTypes.Cluster) error {
+	if _, ok := c.CustomData[singlePoolKey]; ok && len(c.Pools) > 1 {
+		return errors.Errorf("only one pool is allowed to use entire cluster as single-pool. %d pools found", len(c.Pools))
```
wpjunior, line 205: Good catch! @morpheu

provision/kubernetes/nodecontainer.go
morpheu: It's necessary cause generic node-containers are always created for all k8s clusters - even for single pool clusters. There is a test case[0] that shows a DS created with a affinity that never gonna be evaluated on a cluster without Tsuru labels for nodes (eg. on EKS). For single pool cluster mode you should always create a specific entry for the related pool if you want it running - generic entry will not work. 
[0] https://github.com/tsuru/tsuru/pull/2411/files/03a21c146c9994ef0bc3eb2d4b5ea016875c9d5e#diff-9df82fdc12e61e08f6e990b6fa5be696R326

cezarsa: Did you consider not creating the `-all` DS at all? We could do something like:
```
if singlePool {
	 if pool == "" {
		return nil
	}
	affinity = &apiv1.Affinity{}
}
```

I think this would make it easier to debug the cluster because we wouldn't have an extra DaemonSet the would never match any node.

provision/kubernetes/provisioner.go
```diff
@@ -201,6 +201,9 @@ func (p *kubernetesProvisioner) InitializeCluster(c *provTypes.Cluster) error {
 }
 
 func (p *kubernetesProvisioner) ValidateCluster(c *provTypes.Cluster) error {
+	if _, ok := c.CustomData[singlePoolKey]; ok && len(c.Pools) > 1 {
```
cezarsa, line 204: Checking for `len(c.Pools) != 1` would be safer because it wouldn't allow default clusters with a single pool annotation.

morpheu: Why not allow a single cluster working as single pool mode and acting as default cluster for kubernetes provisioner?

morpheu: Humm... sorry, I got it :) You're right.

PR #2412, opened 2020-07-02
fix(provision/kubernetes): fix partial rollback failure that leaves different apps versions after an unsuccessful deploy

nettoclaudio:


provision/servicecommon/actions.go
```diff
@@ -82,19 +85,33 @@ func RunServicePipeline(manager ServiceManager, oldVersion appTypes.AppVersion,
...
+			errors.Add(fmt.Errorf("error rolling back updated service for %s[%s] [version %d]: %+v", args.app.GetName(), processName, args.oldVersion.Version(), err))
 		}
 	}
+	if hasOldVersion {
+		if err := args.manager.CleanupServices(args.app, args.oldVersion, args.preserveVersions); err != nil {
+			errors.Add(fmt.Errorf("error cleaning up services after rollback: %+v", err))
+		}
+	}
+	if errors.ToError() != nil {
+		fmt.Fprintf(w, "some errors occured when rolling back services: %+v", errors)
```
cezarsa, line 113: What about returning this error and using another MultiError to include it in the error returned by the `updateServices` action? It would make it easier for us to find future problems caused by failed rollbacks.

nettoclaudio: Hum... Make sense; Changing here...

PR #2417, opened 2020-07-22
provision/cluster: add metrics for cluster info

wpjunior:
in a multi-cluster environment we need to wisely choose the right cluster before using metrics in singleton grafana instance.

This MR add new metric named `tsuru_cluster_info` that will be useful to use as a grafana variable.

### DIFF
--- a/provision/cluster/metrics.go
+++ b/provision/cluster/metrics.go
@@ -1,0 +1,45 @@
+// Copyright 2020 tsuru authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package cluster
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/tsuru/tsuru/log"
+	"github.com/tsuru/tsuru/servicemanager"
+)
+
+var (
+	desc        = prometheus.NewDesc("tsuru_cluster_info", "Basic information about existing clusters", []string{"provisioner", "name"}, nil)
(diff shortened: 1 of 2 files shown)


provision/cluster/metrics.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/tsuru/tsuru/log"
+	"github.com/tsuru/tsuru/servicemanager"
+)
+
+var (
+	desc        = prometheus.NewDesc("tsuru_cluster_info", "Basic information about existing clusters", []string{"provisioner", "name"}, nil)
+	failureDesc = prometheus.NewDesc("tsuru_cluster_fetch_fail", "indicates whether failed to get clusters", []string{}, nil)
```
nettoclaudio, line 15: Is it a best practice count the fails to obtain other metric? I really don't know (I've no metrics design experience with Prom :/).

provision/cluster/metrics.go
```diff
@@ -0,0 +1,45 @@
...
+
+func (c *clustersMetricCollector) Collect(ch chan<- prometheus.Metric) {
+	clusters, err := servicemanager.Cluster.List()
+	failureValue := float64(0)
+	if err != nil {
+		failureValue = float64(1)
+		log.Errorf("Could not get clusters: %s", err.Error())
+	}
+
+	ch <- prometheus.MustNewConstMetric(failureDesc, prometheus.GaugeValue, failureValue)
```
nettoclaudio, line 36: Should not be a counter instead?

wpjunior: I choose a gauge because it could be 0, whether `ClusterList` becomes operational after a downtime.

PR #2426, opened 2020-08-07
feat(service/proxy): propagate query string to service api

nettoclaudio:


service/endpoint_test.go
```diff
@@ -759,12 +760,16 @@ func (s *S) TestPlans(c *check.C) {
 
 func (s *S) TestEndpointProxy(c *check.C) {
 	handlerTest := func(w http.ResponseWriter, r *http.Request) {
+		fmt.Printf("URL: %#v\n", r.URL)
```
cezarsa, line 763: Left from debugging?

nettoclaudio: yes, sorry :cold_sweat:

PR #2429, opened 2020-08-17
 README: fix git clone instructions

SunkPlane29:
While following the instructions to install tsuru-client from source I found that the `git clone` command was missing the protocol. The command returned an error: `fatal: repository 'github.com/tsuru/tsuru-client' does not exist`. So I added the protocol to the command.


cezarsa:
Thanks!

PR #2447, opened 2020-10-09
api: Replaces request ctx with non-cancelable ctx

cezarsa:


net/context.go
```diff
@@ -6,6 +6,20 @@ package net
 
 import "context"
 
+type baseContextKey struct{}
+
+var baseContextValue baseContextKey
+
+func BaseContext(ctx context.Context) context.Context {
```
wpjunior, line 13: @cezarsa could you rename `BaseContext` to `CancelableParentContext`

PR #2456, opened 2020-11-26
Add new fields to help troubleshoot of pods

wpjunior:
Tsuru will expose new fields of pods:

- Pod restart count
- Pod ready state
- Pod CreationTimestamp


app/app.go
```diff
@@ -209,17 +209,27 @@ func (app *App) Units() ([]provision.Unit, error) {
...
 	result["name"] = app.Name
 	result["platform"] = app.Platform
 	if version := app.GetPlatformVersion(); version != "latest" {
 		result["platform"] = fmt.Sprintf("%s:%s", app.Platform, version)
 	}
+
+	prov, err := app.getProvisioner()
+	if err != nil {
+		errMsgs = append(errMsgs, fmt.Sprintf("unable to get provisioner name: %+v", err))
+		return nil, err
```
cezarsa, line 224: Because of this return the `errMsgs` assignment does nothing. I think it's safe to remove this return.

cezarsa (approved):

PR #2466, opened 2021-02-10
Feat/labels and annotations

pedrokiefer:
Add custom labels and annotations to apps.

### DIFF
--- a/api/app_test.go
+++ b/api/app_test.go
@@ -1691,6 +1691,7 @@ func (s *S) TestUpdateAppWithTagsOnly(c *check.C) {
 	a := app.App{Name: "myapp", Platform: "zend", TeamOwner: s.team.Name}
 	err := app.CreateApp(context.TODO(), &a, s.user)
 	c.Assert(err, check.IsNil)
+
 	token := userWithPermission(c, permission.Permission{
 		Scheme:  permission.PermAppUpdate,
 		Context: permission.Context(permTypes.CtxApp, a.Name),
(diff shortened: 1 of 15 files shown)


api/app_test.go
```diff
@@ -1738,6 +1742,90 @@ func (s *S) TestUpdateAppWithTagsWithoutPermission(c *check.C) {
 	c.Assert(recorder.Code, check.Equals, http.StatusForbidden)
 }
 
+func (s *S) TestUpdateAppWithAnnotations(c *check.C) {
+	a := app.App{
+		Name:      "myapp",
+		Platform:  "zend",
+		TeamOwner: s.team.Name,
+		Metadata: appTypes.Metadata{
+			Annotations: []appTypes.MetadataItem{{Name: "c", Value: "someData"}},
```
wpjunior, line 1751: @pedrokiefer why we can use simple map[string]string for annotations ?

types/app/metadata.go
```diff
@@ -0,0 +1,113 @@
+// Copyright 2018 tsuru authors. All rights reserved.
```
wpjunior, line 1: @pedrokiefer let's use 2021 <3

api/app_test.go
cezarsa: I think that's because the metadata is also stored in mongodb as part of the app. And mongodb support for keys including dots is a mess.

types/app/metadata.go
```diff
@@ -0,0 +1,113 @@
...
+		list = append(list, item)
+	}
+	return list
+}
+
+func hasItem(list []MetadataItem, name string) int {
+	n := -1
+	for i, v := range list {
+		if v.Name == name {
+			n = i
```
arthurcgc, line 100: why not: `return i`?

api/app_test.go
wpjunior: Sure! I forgot this problem, in the newer version of MongoDB it is allowed, however, I prefer to stay safe with older versions.

wpjunior (approved):

cezarsa (approved):

PR #2474, opened 2021-04-01
refactor: support PATCH method on router.AddAll

pedrokiefer:
PATCH method is not currently supported, this leads to services APIs not being able to use it through the services proxy.


wpjunior:
Thanks @pedrokiefer

PR #2476, opened 2021-04-08
Allow configurable timeout for healthcheck passing during deploy

cezarsa:
This PR adds the field `deploy_timeout_seconds` to the health check yaml configuration. I'm open for suggestions on a better name for this field.


wpjunior (approved):
LGTM @cezarsa, thanks!

PR #2479, opened 2021-04-19
refactor: improve env-vars API for easier integration with terraform

pedrokiefer:
EnvVars API is not really terraform friendly, and we have to create a diff locally and remove / update with multiple requests.

This PR lets we push update all variables with a single requests, pruning unneeded variables if requested. The pruning only happens to variables managed by the same service ie. `terraform`, `tsuru-client`, and so on.

What is missing from this PR is a change to the database model, making all variables public, except those marked as `private`.

### DIFF
--- a/api/app.go
+++ b/api/app.go
@@ -984,6 +984,10 @@ func setEnv(w http.ResponseWriter, r *http.Request, t auth.Token) (err error) {
 		msg := "You must provide the list of environment variables"
 		return &errors.HTTP{Code: http.StatusBadRequest, Message: msg}
 	}
+	if e.PruneUnused && e.ManagedBy == "" {
+		msg := "Prune unused requires a managed-by value"
+		return &errors.HTTP{Code: http.StatusBadRequest, Message: msg}
+	}
 	appName := r.URL.Query().Get(":app")
 	a, err := getAppFromContext(appName, r)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


app/app.go
```diff
@@ -1577,30 +1577,47 @@ func (app *App) SetEnvs(setEnvs bind.SetEnvArgs) error {
...
 	if setEnvs.Writer != nil {
 		fmt.Fprintf(setEnvs.Writer, "---- Setting %d new environment variables ----\n", len(setEnvs.Envs))
 	}
+
+	if setEnvs.PruneUnused {
+		for name, value := range app.Env {
+			ok := envInSet(name, setEnvs.Envs)
+			// only prune variables managed by requested
+			if !ok && value.ManagedBy == setEnvs.ManagedBy {
+				delete(app.Env, name)
```
wpjunior, line 1597: I suggest to we have a message in streaming to inform that some environment variable will be deleted. @pedrokiefer

api/app.go
```diff
@@ -996,11 +996,13 @@ func setEnv(w http.ResponseWriter, r *http.Request, t auth.Token) (err error) {
 		return permission.ErrUnauthorized
 	}
 	var toExclude []string
```
wpjunior, line 998: @pedrokiefer I want to people use `PruneUnused` in pair of `ManagedBy`, could we have a validation to avoid use of `PruneUnused` without providing a `ManagedBy` ?

wpjunior:
Thanks @pedrokiefer!

PR #2490, opened 2021-06-21
Feat backend config

arthurcgc:


provision/kubernetes/deploy.go
```diff
@@ -433,17 +436,88 @@ func validateHC(hc *provTypes.TsuruYamlHealthcheck, client *ClusterClient) error
...
 	}
 
 	return nil
 }
 
-func probesFromHC(hc *provTypes.TsuruYamlHealthcheck, client *ClusterClient, port int) (hcResult, error) {
+func int64PointerFromInt(v int) *int64 {
+	p := new(int64)
+	*p = int64(v)
+	return p
```
cezarsa, line 452:
```suggestion
	p := int64(v)
	return &p
```

provision/kubernetes/deploy.go
```diff
@@ -59,6 +60,8 @@ const (
 	buildIntercontainerDone = buildIntercontainerPath + "/done"
 	defaultHttpPortName     = "http-default"
 	defaultUdpPortName      = "udp-default"
+	backendConfigName       = "backendconfigs.cloud.google.com"
```
cezarsa, line 63:
```suggestion
	backendConfigCRDName    = "backendconfigs.cloud.google.com"
```

provision/kubernetes/deploy.go
```diff
@@ -433,17 +436,88 @@ func validateHC(hc *provTypes.TsuruYamlHealthcheck, client *ClusterClient) error
 		if hc.TimeoutSeconds >= hc.IntervalSeconds {
 			hc.IntervalSeconds = hc.TimeoutSeconds + 1
 		}
+
+		if err := ensureBackendConfig(ctx, client, app, hc); err != nil {
```
cezarsa, line 440: In my opinion this call shouldn't be here. A good place for it might be right before ensuring services exist in: https://github.com/tsuru/tsuru/blob/6dd56808976bd8ee00dd35eac5a26f4dea288405/provision/kubernetes/deploy.go#L1243

It's okay to call `version.TsuruYamlData()` again if needed.

provision/kubernetes/deploy.go
```diff
@@ -433,17 +436,88 @@ func validateHC(hc *provTypes.TsuruYamlHealthcheck, client *ClusterClient) error
...
+func ensureBackendConfig(ctx context.Context, client *ClusterClient, a provision.App, hc *provTypes.TsuruYamlHealthcheck) error {
+	exists, err := backendConfigCRDExists(ctx, client)
+	if err != nil {
+		return err
+	}
+	if !exists {
+		return nil
+	}
+
+	backendConfigName := fmt.Sprintf("%s-backend-config", a.GetName())
```
cezarsa, line 464: We should create a helper function for naming backend config similar to those in https://github.com/tsuru/tsuru/blob/6dd56808976bd8ee00dd35eac5a26f4dea288405/provision/kubernetes/helpers.go#L107-L110

provision/kubernetes/deploy.go
```diff
@@ -433,17 +436,88 @@ func validateHC(hc *provTypes.TsuruYamlHealthcheck, client *ClusterClient) error
...
 }
 
-func probesFromHC(hc *provTypes.TsuruYamlHealthcheck, client *ClusterClient, port int) (hcResult, error) {
+func int64PointerFromInt(v int) *int64 {
+	p := new(int64)
+	*p = int64(v)
+	return p
+}
+
+func ensureBackendConfig(ctx context.Context, client *ClusterClient, a provision.App, hc *provTypes.TsuruYamlHealthcheck) error {
```
cezarsa, line 455: I think it would be nice for organization if we moved these functions related to backend config to a new file. Something like `gcp_backendconfig.go`.

provision/kubernetes/deploy.go
```diff
@@ -433,17 +436,88 @@ func validateHC(hc *provTypes.TsuruYamlHealthcheck, client *ClusterClient) error
...
+	backendConfig := &backendconfigv1.BackendConfig{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      backendConfigName,
+			Namespace: ns,
+		},
+		Spec: backendconfigv1.BackendConfigSpec{
+			HealthCheck: &backendconfigv1.HealthCheckConfig{
+				CheckIntervalSec:   intervalSec,
+				TimeoutSec:         timeoutSec,
+				HealthyThreshold:   healthyThreshold,
```
cezarsa, line 487: It's safer to hardcode this as `1` for now since we don't have a specific flag for setting it on regular probes and the default value there is one.

provision/kubernetes/deploy.go
```diff
@@ -1413,6 +1502,11 @@ func (m *serviceManager) ensureServices(ctx context.Context, a provision.App, pr
 		if err != nil {
 			return errors.WithMessage(err, "could not to parse base services annotations")
 		}
+		if annotations == nil {
+			annotations = make(map[string]string)
+		}
+		annotations[backendConfigKey] = fmt.Sprintf("{\"default\":\"%s\"}", backendConfigName)
```
cezarsa, line 1508: This value cannot be a constant. It must be the name of the backend config created for this app/process. The name helper function I mentioned above can be useful here as well.

provision/kubernetes/deploy.go
```diff
@@ -1413,6 +1417,11 @@ func (m *serviceManager) ensureServices(ctx context.Context, a provision.App, pr
 		if err != nil {
 			return errors.WithMessage(err, "could not to parse base services annotations")
 		}
+		if annotations == nil {
+			annotations = make(map[string]string)
+		}
+		annotations[backendConfigKey] = fmt.Sprintf("{\"default\":\"%s\"}", backendConfigNameForApp(a, process))
```
wpjunior, line 1423: @arthurcgc we could inject this `backendConfigKey` just when backendConfig CRD exists, right ?

arthurcgc: @wpjunior I mean we could... but it would mean another call to the k8s api just to set the annotation... what do you think @cezarsa ?

provision/kubernetes/gcp_backendconfig.go
```diff
@@ -0,0 +1,97 @@
...
+
+	existingBackendConfig, err := cli.CloudV1().BackendConfigs(ns).Get(ctx, backendConfigName, metav1.GetOptions{})
+	if k8sErrors.IsNotFound(err) {
+		existingBackendConfig = nil
+	} else if err != nil {
+		return err
+	}
+
+	if existingBackendConfig != nil {
+		backendConfig.ResourceVersion = existingBackendConfig.ResourceVersion
```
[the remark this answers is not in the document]
arthurcgc: The update function already takes care of that, doesn't it?

provision/kubernetes/deploy.go
wpjunior: the output of `ensureBackendConfig` could return this @arthurcgc

wpjunior:
Thanks @arthurcgc

PR #2497, opened 2021-07-20
Set default plan based on pool constraints

morpheu:
Use first match plan as default when plan pool constraint is set

### DIFF
--- a/provision/pool/pool.go
+++ b/provision/pool/pool.go
@@ -157,40 +157,39 @@ func (p *Pool) GetPlans() ([]string, error) {
 	return nil, ErrPoolHasNoPlan
 }
 
-func (p *Pool) GetDefaultPlan() (string, error) {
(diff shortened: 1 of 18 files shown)


provision/pool/pool.go
```diff
@@ -157,6 +157,42 @@ func (p *Pool) GetPlans() ([]string, error) {
...
+		allowed, err = p.allowedValues()
+		if err != nil {
+			return "", err
+		}
+		if len(allowed[ConstraintTypePlan]) == 1 {
+			return allowed[ConstraintTypePlan][0], nil
+		}
+		return defaultPlan.Name, nil
+	}
+	plans, err := servicemanager.Plan.List(p.ctx)
```
wpjunior, line 184: @morpheu could we use Plan.FindByName ?

morpheu: Sure, I did a little refactor :) Thanks

wpjunior (approved):
LGTM @morpheu

PR #2498, opened 2021-07-20
refactor: avoid unnecessary memory allocation on interface checks

LucasMendesl:
improvment to avoid unnecessary memory allocation when compiler checks interface implementation

### DIFF
--- a/provision/kubernetes/pkg/client/clientset/versioned/fake/clientset_generated.go
+++ b/provision/kubernetes/pkg/client/clientset/versioned/fake/clientset_generated.go
@@ -62,7 +62,7 @@ func (c *Clientset) Tracker() testing.ObjectTracker {
 	return c.tracker
 }
 
-var _ clientset.Interface = &Clientset{}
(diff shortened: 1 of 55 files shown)


provision/kubernetes/pkg/client/clientset/versioned/fake/clientset_generated.go
```diff
@@ -62,7 +62,7 @@ func (c *Clientset) Tracker() testing.ObjectTracker {
 	return c.tracker
 }
 
-var _ clientset.Interface = &Clientset{}
+var _ clientset.Interface = (*Clientset)(nil)
```
cezarsa, line 65: This code is automatically generated by kubernetes scripts. It should not be changed.

cezarsa (commented):
Hi @LucasMendesl, first of all thanks for your contribution. I have no objections against this change even though I think the memory impact is negligible, if any. Would you mind addressing the inline comments about generated files?

PR #2499, opened 2021-07-21
Fix: app stop/start

arthurcgc:
- This PR fixes a bug when using kubernetes provider and `app start`. It would not spawn any units when calling `app start`.
- This PR makes it so we don't delete the apps k8s services when running `app stop`


app/app.go
```diff
@@ -1308,6 +1313,34 @@ func (app *App) Stop(ctx context.Context, w io.Writer, process, versionStr strin
 	if err != nil {
 		return err
 	}
+	annotationSuffix := process
```
wpjunior, line 1316: @arthurcgc could you extract that new block to a new function ?

app/app.go
```diff
@@ -1308,6 +1313,34 @@ func (app *App) Stop(ctx context.Context, w io.Writer, process, versionStr strin
...
+	}
+	units, err := app.Units()
+	if err != nil {
+		return err
+	}
+	app.Metadata.Annotations = append(app.Metadata.Annotations, appTypes.MetadataItem{
+		Name:  appTypes.PastUnitsAnnotationPrefix + annotationSuffix,
+		Value: strconv.Itoa(len(units)),
+	})
+	upArgs := UpdateAppArgs{
```
wpjunior, line 1328: it is possible to do it without restart ? @arthurcgc

provision/kubernetes/deploy.go
```diff
@@ -802,7 +802,7 @@ type serviceManager struct {
 
 var _ servicecommon.ServiceManager = &serviceManager{}
 
-func (m *serviceManager) CleanupServices(ctx context.Context, a provision.App, deployedVersion int, preserveOldVersions bool) error {
+func (m *serviceManager) CleanupServices(ctx context.Context, a provision.App, deployedVersion int, preserveOldVersions bool, newVersionSpec servicecommon.ProcessSpec) error {
 	depGroups, err := deploymentsDataForApp(ctx, m.client, a)
```
wpjunior, line 806: @arthurcgc could we avoid call of `CleanupServices` using stop ?, stop operation should be a atomic update in replicas field.

app/app.go
```diff
@@ -2128,7 +2128,10 @@ func (app *App) Start(ctx context.Context, w io.Writer, process, versionStr stri
 	if err != nil {
 		return err
 	}
-	err = prov.Start(ctx, app, process, version)
+	if err != nil {
```
wpjunior, line 2131: @arthurcgc we have two `if err != nil`

wpjunior:
Thanks @arthurcgc

PR #2502, opened 2021-08-05
Fix app shell isolated

arthurcgc:
This PR makes the app shell isolated container declare it's resource requirements beforehand.
The resource requirements are based from the app plan.

### DIFF
--- a/provision/kubernetes/provisioner.go
+++ b/provision/kubernetes/provisioner.go
@@ -1587,7 +1587,9 @@ func (p *kubernetesProvisioner) ExecuteCommand(ctx context.Context, opts provisi
 		termSize: size,
 		tty:      opts.Stdin != nil,
 	}
-	if len(opts.Units) == 0 {
+
+	isIsolated := len(opts.Units) == 0
+	if isIsolated {
 		return runIsolatedCmdPod(ctx, client, eOpts)
 	}
 	for _, u := range opts.Units {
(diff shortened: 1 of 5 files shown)


wpjunior:
LGTM @arthurcgc 🚀

provision/kubernetes/provisioner.go
```diff
@@ -1445,6 +1447,17 @@ func runIsolatedCmdPod(ctx context.Context, client *ClusterClient, opts execOpts
...
+	var requirements *apiv1.ResourceList
+	memory := opts.app.GetMemory()
+	cpu := opts.app.GetMilliCPU()
+	if memory > 0 && cpu > 0 {
+		requirements = &apiv1.ResourceList{
+			apiv1.ResourceMemory: *resource.NewQuantity(memory, resource.BinarySI),
+			apiv1.ResourceCPU:    *resource.NewMilliQuantity(int64(cpu), resource.DecimalSI),
+		}
+	}
+
```
nettoclaudio, line 1460: I'd replace this code block in favor of the snippet below. Actually, you could extract it to a helper function and reuse the same logic to create the resource list in both places, WDYT?

https://github.com/tsuru/tsuru/blob/80db3599179e4fdc7e0d9876756342dce4f86e93/provision/kubernetes/deploy.go#L654-L677

nettoclaudio: You could move this whole implementation to `helpers.go` file though.

nettoclaudio (changes requested):
We can reuse the code to create the resource list from deploy stages, PTAL.

PR #2504, opened 2021-08-05
feat(app/shell): fallback to dash when bash is not found

nettoclaudio:
Today we cannot run app shell command against an app that doesn't have `bash`, such as Alpine-based platforms. This PR tries to find `bash` executable before executing it or calls `dash` otherwise.


wpjunior:
Thanks @nettoclaudio

PR #2517, opened 2021-10-05
fix typos in doc

garanews:


wpjunior:
Thanks so very much @garanews =)

PR #2525, opened 2021-11-17
feat(provision/kubernetes): support PDB for each Tsuru process

nettoclaudio:
This PR adds support to set pod disruption budget ([reference](https://kubernetes.io/docs/tasks/run-application/configure-pdb/)) when using an app within a Kubernetes pool.

As described below, it provides several ways to set the minimal available of Tsuru process - last items has higher precedence over the first ones:

1) no min available is set (default behavior);
2) cluster administrator can set a cluster-wide min available value;
3) cluster administrator can set a pool-wide min available value;
4) app's developer can set the minimum available on PDB by creating an autoscale for the process such min of replicas on HPA will be the minimum available on PDB;

Example:

By creating an autoscale for "web" process with 10 replicas at least,  it's going to configure a pod disruption budget with 10 min available replicas PDB for "web" process as well.
```
tsuru unit autoscale set -a my-app -p web --cpu 90% --min 10 --max 1000 
```


provision/kubernetes/deploy.go
```diff
@@ -862,6 +862,24 @@ func (m *serviceManager) CleanupServices(ctx context.Context, a provision.App, d
...
+		multiErrors.Add(err)
+	}
+	for _, pdb := range pdbs {
+		labels := labelSetFromMeta(&pdb.ObjectMeta)
+		process := labels.AppProcess()
+		if _, toKeep := processInUse[process]; toKeep {
+			continue
+		}
+
+		fmt.Fprintf(m.writer, " ---> Cleaning up PDB %s\n", pdb.Name)
```
wpjunior, line 876:
```suggestion
		fmt.Fprintf(m.writer, " ---> Cleaning up PodDirruptionBudget %s\n", pdb.Name)
```

PR #2526, opened 2021-11-17
Avoid creation of versioned services during app deploy

arthurcgc:


provision/kubernetes/cluster.go
```diff
@@ -498,6 +500,17 @@ func (c *ClusterClient) SinglePool() (bool, error) {
 	return strconv.ParseBool(singlePool)
 }
 
+func (c *ClusterClient) AvoidMultipleServices() (bool, error) {
```
wpjunior, line 503: @arthurcgc I did not see a call to `AvoidMultipleServices` bellow, I think that we could use that without flag.

arthurcgc: That's actually a really nice catch... The right thing to do would be to check if `newVersion` and the clusterFlag is set inside ensureServices... do you agree?

wpjunior: yeah @arthurcgc, in the feature we could drop this flag, them turn this a default behavior

provision/kubernetes/deploy.go
```diff
@@ -1415,12 +1415,21 @@ func (m *serviceManager) ensureServices(ctx context.Context, a provision.App, pr
...
-			ports:    svcPorts,
-		})
+		shouldAvoidServiceVersion := false
+		shouldAvoidServiceVersion, err = m.client.AvoidMultipleServices()
+		if err != nil {
+			return err
+		}
+		// only add new services if a new version is specified
+		// if no new version is specified we only create the default service and the headless service
+		if newVersion && !shouldAvoidServiceVersion {
```
nettoclaudio, line 1425: I think it should be ORed.

provision/kubernetes/cluster.go
nettoclaudio, line 503: What about `createAdditionalServiceForVersion`?
```suggestion
func (c *ClusterClient) createAdditionalServiceForVersion() (bool, error) {
```

nettoclaudio (changes requested):

provision/kubernetes/cluster.go
```diff
@@ -99,6 +100,7 @@ var (
 		buildServiceAccountKey:        "Custom service account used in build containers.",
 		disablePlatformBuildKey:       "Disable platform image build in cluster.",
 		sidecarRegistryKey:            "Override for deploy sidecar image registry.",
+		versionedServices:             "Allow the creation of multiple services for each pair of {process, version} of the app",
```
wpjunior, line 103: @arthurcgc I think that we could describe of this opt,  like this:
```
Always create of multiple services for each pair of {process, version} of the app, the default behavior creates only in multi version deployments.
```

PR #2534, opened 2022-02-11
Set PDB using 90% of HPA minReplicas or 10% maxUnavailable

morpheu:
Change how PDB is calculated using floor(90% of HPA minReplicas) or 10% maxUnavailable


nettoclaudio (approved):
lgtm

PR #2537, opened 2022-03-14
Fix override old versions

arthurcgc:


nettoclaudio (changes requested):

provision/kubernetes/deploy.go
```diff
@@ -1254,6 +1254,21 @@ func (m *serviceManager) DeployService(ctx context.Context, opts servicecommon.D
...
+		var deps *appsv1.DeploymentList
+		processSelector := fmt.Sprintf("tsuru.io/app-name=%s, tsuru.io/app-process=%s", opts.App.GetName(), opts.ProcessName)
+		deps, err = m.client.AppsV1().Deployments(ns).List(ctx, metav1.ListOptions{
+			LabelSelector: processSelector,
+		})
+		if err != nil && !k8sErrors.IsNotFound(err) {
+			return errors.WithStack(err)
+		}
+		if deps.Items != nil {
+			opts.Replicas = len(deps.Items)
```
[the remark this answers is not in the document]
arthurcgc: SGTM

provision/kubernetes/deploy.go
```diff
@@ -1254,6 +1254,27 @@ func (m *serviceManager) DeployService(ctx context.Context, opts servicecommon.D
...
+	if opts.OverrideVersions {
+		var deps *appsv1.DeploymentList
+		processSelector := fmt.Sprintf("tsuru.io/app-name=%s, tsuru.io/app-process=%s, tsuru.io/is-routable=true", opts.App.GetName(), opts.ProcessName)
+		deps, err = m.client.AppsV1().Deployments(ns).List(ctx, metav1.ListOptions{
+			LabelSelector: processSelector,
+		})
+		if err != nil && !k8sErrors.IsNotFound(err) {
+			return errors.WithStack(err)
+		}
+		if deps.Items != nil {
```
nettoclaudio, line 1267: Useless nil checking... If it's nil it doesn't enter in the range loop.

provision/kubernetes/deploy.go
```diff
@@ -1254,6 +1254,27 @@ func (m *serviceManager) DeployService(ctx context.Context, opts servicecommon.D
...
+			totalReplicas := 0
+			for _, dep := range deps.Items {
+				if dep.Spec.Replicas != nil {
+					totalReplicas += int(*dep.Spec.Replicas)
```
nettoclaudio, line 1271: I know it might not happen but just to avoid an unexpected nil pointer at runtime, what about to check if replicas is set before adding it.

```suggestion
					if r := dep.Spec.Replicas; r != nil {
						totalReplicas += int(*r)
					}
```

provision/kubernetes/deploy.go
```diff
@@ -1254,6 +1254,27 @@ func (m *serviceManager) DeployService(ctx context.Context, opts servicecommon.D
 	if err != nil {
 		return errors.WithStack(err)
 	}
+
+	if opts.OverrideVersions {
+		var deps *appsv1.DeploymentList
```
nettoclaudio, line 1259: It's more safe defining it as a regular variable instead of a pointer.
```suggestion
		var deps appsv1.DeploymentList
```

nettoclaudio (changes requested):

PR #2543, opened 2022-04-05
event: add check for extra event fields on event block

morpheu:
Add option to check and block for extra fields (name=value) on event

### DIFF
--- a/event/block.go
+++ b/event/block.go
@@ -39,14 +39,15 @@ func (e ErrEventBlocked) Error() string {
 }
 
 type Block struct {
-	ID        bson.ObjectId `bson:"_id,omitempty"`
-	StartTime time.Time
-	EndTime   time.Time `bson:"endtime,omitempty"`
-	KindName  string
-	OwnerName string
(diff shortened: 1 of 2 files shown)


event/block.go
```diff
@@ -39,14 +39,15 @@ func (e ErrEventBlocked) Error() string {
...
-	Target    Target `bson:"target,omitempty"`
-	Reason    string
-	Active    bool
+	ID          bson.ObjectId `bson:"_id,omitempty"`
+	StartTime   time.Time
+	EndTime     time.Time `bson:"endtime,omitempty"`
+	KindName    string
+	OwnerName   string
+	Target      Target            `bson:"target,omitempty"`
+	ExtraFields map[string]string `bson:"extraFields,omitempty"`
```
wpjunior, line 48: @morpheu could you Rename ExtraFields to Restrictions or Conditions ?

wpjunior (changes requested):

event/block.go
morpheu: Sure! Done :)

PR #2544, opened 2022-04-06
Save and retrieve unit status from storage

arthurcgc:
This PR reintroduces a feature removed in [#2539](https://github.com/tsuru/tsuru/pull/2539), tsuru-api persists the last state of the process-version pair inside its storage provider and tries to restore it when attempting to start a previously stopped deployment.

TLDR: if `tsuru app stop -a <appname> --process <processname> --version <versionnumber>` had 3 units running, a `tsuru app start -a <appname> --process <processname> --version <versionnumber>` will restore it with 3 units

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -1392,6 +1392,7 @@ func (app *App) Restart(ctx context.Context, process, versionStr string, w io.Wr
 	return nil
 }
 
+// vpPair represents each version-process pair
 type vpPair struct {
 	version int
 	process string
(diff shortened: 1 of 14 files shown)


app/app.go
```diff
@@ -1392,6 +1392,80 @@ func (app *App) Restart(ctx context.Context, process, versionStr string, w io.Wr
...
 
+// vpPair represents each version-process pair
+type vpPair struct {
+	version int
+	process string
+}
+
+func cleanupOtherProcesses(vpMap map[vpPair]int, process string) {
+	for pair := range vpMap {
+		if strings.Compare(pair.process, process) != 0 {
```
wpjunior, line 1403: why not use par.process != process ?

PR #2548, opened 2022-05-23
provision/kubernetes/deploy: custom annotations/labels on deploy stage

morpheu:
Also set user custom annotations/labels on deploy stage


wpjunior:
Thanks @morpheu

PR #2558, opened 2022-07-28
fix: check for empty value during role-assign

arthurcgc:


api/permission.go
```diff
@@ -317,11 +331,24 @@ func assignRole(w http.ResponseWriter, r *http.Request, t auth.Token) (err error
 	defer func() { evt.Done(err) }()
 	email := InputValue(r, "email")
 	contextValue := InputValue(r, "context")
+	contextType := InputValue(r, "contextType")
+	if contextType == "" {
+		contextType = "global"
+	}
```
nettoclaudio, line 337: Sorry... I'm a bit confused with these changes, can you clarify for me? :grimacing: 

Since `tsuru` CLI doesn't set the param `contextType` to user/token/group role assign endpoints today. Then, at least for old client ones, this code will always assume the default value `global`. However, if I understood your logic correctly, it will return an error when the role doesn't match its context type with `global`, isn't it?

PR #2564, opened 2022-08-19
allow use same cname on different apps

morpheu:


app/app.go
```diff
@@ -2375,6 +2375,20 @@ func (app *App) AddRouter(appRouter appTypes.AppRouter) error {
...
+	if err != nil {
+		return err
+	}
+	defer conn.Close()
+	err = conn.Apps().Find(bson.M{"cname": bson.M{"$in": cnames}, "name": bson.M{"$ne": app.Name}, "routers": appRouter}).One(&appCName)
+	if err != nil && err != mgo.ErrNotFound {
+		return err
+	}
+	if appCName.Name != "" {
+		return errors.New(fmt.Sprintf("cname %s already exists for app %s using router %s", appCName.CName[0], appCName.Name, appRouter.Name))
```
tcarreira, line 2390: technically, the cname already existing `appCName.CName[0]` could be different from the one present in the `app.GetCname()` 

eg: 
```
app.GetCname() = ["awesome-domain", "fantastic-domain"]
appCName.GetCname() = ["bad-bad-domain", "fantastic-domain"]
```
would get the error "cname `bad-bad-domain` already exists for app xxx using router yyy", for which I would react with a "What The Flaaaag?"

Maybe adding this case to the tests would be nice?

morpheu: You're right. I totally forgot to proper change this error message and return the right duplicated CName :(

tcarreira (approved):

PR #2565, opened 2022-08-23
feat(auth/token): support creating app from team token

nettoclaudio:
This PR allows creating an app from a team token as required for apps managed by external tools like Terraform.

### DIFF
--- a/app/app_test.go
+++ b/app/app_test.go
@@ -678,9 +678,13 @@ func (s *S) TestCreateAppProvisionerFailures(c *check.C) {
 }
 
 func (s *S) TestCreateAppUserFromTsuruToken(c *check.C) {
+	s.mockService.UserQuota.OnInc = func(item quota.QuotaItem, delta int) error {
+		return stderrors.New("cannot be called")
+	}
 	user := auth.User{
-		Email: "my-token@token.tsuru.internal",
-		Quota: quota.UnlimitedQuota,
+		Email:     "my-token@token.tsuru.invalid",
+		Quota:     quota.UnlimitedQuota,
+		FromToken: true,
 	}
 	a := App{
 		Name:      "my-app",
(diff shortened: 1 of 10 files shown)


auth/team_token.go
```diff
@@ -8,17 +8,32 @@ import (
...
+// TsuruTokenEmailDomain is the e-mail domain used to fake users from a team
+// token.
+//
+// NOTE: Following the RFC2606, the top-level domain ".invalid" is reserved,
+// so regular Tsuru users should not be able to register using it. See more:
+// https://www.rfc-editor.org/rfc/rfc2606
+const TsuruTokenEmailDomain = "token.tsuru.invalid"
+
+func IsEmailFromTeamToken(email string) bool {
+	return strings.HasSuffix(email, fmt.Sprintf("@%s", TsuruTokenEmailDomain))
```
tcarreira, line 34: is it possible to arrive at this line with a url-encoded email? 
eg: `example%40token.tsuru.invalid`.
This would return false, although it should return true

app/app_test.go
```diff
@@ -677,6 +677,20 @@ func (s *S) TestCreateAppProvisionerFailures(c *check.C) {
 	c.Assert(err, check.NotNil)
 }
 
+func (s *S) TestCreateAppUserFromTsuruToken(c *check.C) {
+	user := auth.User{
+		Email: "my-token@token.tsuru.internal",
```
tcarreira, line 682: Shouldn't this email end with `token.tsuru.invalid` in order to get true from `IsEmailFromTeamToken()`?

nettoclaudio: Sure, I overlooked it.

auth/team_token.go
nettoclaudio: I don't think so :crossed_fingers:

tcarreira (approved):

morpheu (approved):

tcarreira (approved):
👌

PR #2580, opened 2022-10-17
fix: RecordingFs{}.Rename dirs (and its children)

### LINKED ISSUE #2579: RecordingFs.Rename() does not work for directories content
**Description**
> A clear and concise description of what the bug is.

The RecordingFs.Rename() only works for specific files.
When trying to rename a directory (expecting its content to be available on the newer path), only the directory is renamed. Every file inside it will remain on the old path.

**Steps to reproduce**

```
func main()
	fs := fstest.RecordingFs{}
	fs.Mkdir("/tmp/dir1", 0755)
	fs.Create("/tmp/dir1/file1")
	fs.Create("/tmp/dir1/file2")
	for _, fname := range []string{"/tmp/dir1", "/tmp/dir1/file1", "/tmp/dir1/file2"} {
		_, err := fs.Stat(fname)
		assert(err == nil, fmt.Sprintf("%s should exist", fname))
	}

	fs.Rename("/tmp/dir1", "/tmp/dir2")
	for _, fname := range []string{"/tmp/dir2", "/tmp/dir2/file1", "/tmp/dir2/file2"} {
		_, err := fs.Stat(fname)
		assert(err == nil, fmt.Sprintf("%s should exist", fname))
	}
}

// go run main.go
//   assert failed! /tmp/dir2/file1 should exist
//   assert failed! /tmp/dir2/file2 should exist
```

**Expected result**
> A clear and concise description of what you expected to happen.

When renaming (moving) a directory, every child should me renamed (moved) as well

tcarreira:
closes #2579

### DIFF
--- a/fs/fstest/testing_fs.go
+++ b/fs/fstest/testing_fs.go
@@ -289,6 +289,22 @@ func (r *RecordingFs) Rename(oldname, newname string) error {
 	if r.files == nil {
 		r.files = make(map[string]*FakeFile)
 	}
+	if strings.HasPrefix(newname, oldname) { // moving into itself
+		return os.ErrInvalid
+	}
+
+	// rename all dir's childs (very non-performant)
+	if r.files[oldname] != nil && r.files[oldname].dir {
+		oldDirName := oldname + string(os.PathSeparator)
+		for fname := range r.files {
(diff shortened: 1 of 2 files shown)


nettoclaudio (approved):
lgtm

PR #2581, opened 2022-10-18
Enable alias on register command

### LINKED ISSUE #2576: Tsuru CLI - alias for commands
**Is your feature request related to a problem? Please describe.**
N/A

**Describe the solution you'd like**
> Be able to run tsuru cli commands with a alias eg. `tsuru log` same with `tsuru logs`

**Describe alternatives you've considered**.                                                                   
> Create a alias on bash.

**Additional context**
> I'm used to use `logs` as command for log view, like in `docker logs`. It would be awesome being able to call with `log` or `logs`

RaphaelVRossi:
Close #2576 

Enable alias on register new command.

When register a new command, now you can inform `alias` field to generate a new alias.

e.g. `tsuru app log` and `tsuru app logs` call the same command run.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -151,6 +151,14 @@ func (m *Manager) Register(command Command) {
 		m.Commands = make(map[string]Command)
 	}
 	name := command.Info().Name
+	m.RegisterCommandByName(name, command)
+	alias := command.Info().Alias
+	if alias != "" {
+		m.RegisterCommandByName(alias, command)
+	}
+}
+
+func (m *Manager) RegisterCommandByName(name string, command Command) {
 	_, found := m.Commands[name]
 	if found {
 		panic(fmt.Sprintf("command already registered: %s", name))
@@ -621,6 +629,7 @@ func (c *Context) RawOutput() {
 
 type Info struct {
 	Name    string
+	Alias   string
 	MinArgs int
 	MaxArgs int
 	Usage   string
--- a/cmd/cmd_test.go
+++ b/cmd/cmd_test.go
@@ -56,6 +56,12 @@ func (s *S) TestRegister(c *check.C) {
 	c.Assert(badCall, check.PanicMatches, "command already registered: foo")
 }
 
+func (s *S) TestRegisterCommandByName(c *check.C) {
+	globalManager.RegisterCommandByName("foobar", &TestCommand{})
+	badCall := func() { globalManager.RegisterCommandByName("foobar", &TestCommand{}) }
+	c.Assert(badCall, check.PanicMatches, "command already registered: foo")
+}
+
 func (s *S) TestRegisterDeprecated(c *check.C) {
 	originalCmd := &TestCommand{}
 	globalManager.RegisterDeprecated(originalCmd, "bar")
@@ -107,6 +113,7 @@ func (s *S) TestRegisterTopicMultiple(c *check.C) {
 
 type TopicCommand struct {
 	name     string
+	alias    string
 	executed bool
 	args     []string
 }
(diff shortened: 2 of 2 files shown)


cmd/cmd.go
```diff
@@ -151,6 +151,14 @@ func (m *Manager) Register(command Command) {
 		m.Commands = make(map[string]Command)
 	}
 	name := command.Info().Name
+	m.RegisterCommandByName(name, command)
+	alias := command.Info().Alias
+	if alias != "" {
+		m.RegisterCommandByName(alias, command)
+	}
```
nettoclaudio, line 158: I'm not 100% sure but it could print the command info twice on help message - in different lines, though.

I'd take a better look on how we produce that help message and I'd show the comand + alias in same line - separeted by comma. For example:
```
app log, app logs                    Shows log entries for an application
```

Or maybe register the alias as a hidden command so it won't be generated on the help message even though it works fine when you fire on the command line.

RaphaelVRossi: You are correct, with this changes, help message show twice `app log`

```sh
./bin/tsuru app
App is a program source code running on Tsuru

The following commands are available in the "app" topic:

[...]
  app log                    Shows log entries for an application
  app log                    Shows log entries for an application
[...]
```

I tried to change where is printed help message, but didn't find.

So I didn't append command alias inside `m.topicCommand`, with that change, only show once at help message:

```sh
./bin/tsuru app
App is a program source code running on Tsuru

The following commands are available in the "app" topic:

  [...]
  app list                   Lists all apps that you have access to
  app log                    Shows log entries for an application
  app metadata get           Retrieves metadata for an application
  [...]
```

What do you think @nettoclaudio ?

RaphaelVRossi: My last comment was about when you call `tsuru app` to show all available commands.

But this other test, I only call `tsuru` and the real command and the alias was in the list, is that correct?!

```sh 
./bin/tsuru
Client version: dev.

Usage: tsuru command [args]

Available commands:
  [...]
  app list                   Lists all apps that you have access to
  app log                    Shows log entries for an application
  app logs                   Shows log entries for an application
  app metadata get           Retrieves metadata for an application
  [...]
```

nettoclaudio: I'm OK on this approach.

To be honest, I don't like to maintain this code on our side anymore. There are a lot of better CLI modules (e.g. spf13/cobra, urfave/cli and so on) than ours. I'm in favor to use them and remove this bunch of code. I don't know if all Tsuru maintainers agree, just my opnion.

nettoclaudio:
Sorry for taking so long to answer. I was trying to write down the aliases at the same line as the main command, I thought we can do it this way:

You should change the [`dumpCommands`](https://github.com/tsuru/tsuru/blob/main/cmd/cmd.go#L482) to only call the `formatDescriptionLine` function on the main command. Something like this:

```golang
func (m *Manager) dumpCommands(commands []string) string {
// ...
		for _, command := range commands {
				// check if command is an alias, if so skip
				info := m.Commands[command].Info()
				output += formatDescriptionLine(command, info.Alias, info.Description.Desc, maxCmdSize)
		}
// ...
```

Then, you should need to change the [`formatDescriptionLine`](https://github.com/tsuru/tsuru/blob/abc5e18fa17361c4279356d1874f4938d0b53557/cmd/cmd.go#L482) to receive the alias as new function argument and print them.

```golang
func formatDescriptionLine(label, alias, description string, maxSize int) string {
// ...
				cmds := strings.Join([]string{label, alias}, ", ")
				return fmt.Sprintf(fmtStr, strings.ReplaceAll(cmds, "-", " "), description)
}
```

It should produce something like:
```
  app log, app logs          Shows log entries for an application
```

RaphaelVRossi:
Thank you @nettoclaudio! I will work on that.

PR #2594, opened 2022-12-21
fix(k8s): app update should not create unit of zeroed process

nettoclaudio:
Fixing a bug where an app update creates an unit of base version even it wasn't running before.

Steps to reproduce:
1. App with two versions running simultaneously;
1.1. Base version w/ zero units (`tsuru unit remove ... <all units>`);
1.2. Extra version running regardless the numbers of units; 
2.  Run an app update (allowing restart);
3. A new units of base version gets running;


arthurcgc (approved):
lgtm

PR #2604, opened 2023-03-20
API to bind Job on a service

wpjunior:


api/event_test.go
```diff
@@ -82,23 +82,23 @@ func (s *EventSuite) generateEventCustomData(appName string) *app.DeployOptions
...
-			Env: map[string]bind.EnvVar{
+			Env: map[string]appTypes.EnvVar{
 				"MY_PASSWORD": {
 					Name:   "MY_PASSWORD",
 					Public: false,
 					Value:  "password123",
 				},
 			},
-			ServiceEnvs: []bind.ServiceEnvVar{
+			ServiceEnvs: []appTypes.ServiceEnvVar{
```
arthurcgc, line 92: I think we could take the opportunity to move EnvVar and ServiceEnvVar to a more generic package like `types/`

provision/pool/pool.go
```diff
@@ -605,6 +605,31 @@ func GetPoolByName(ctx context.Context, name string) (*Pool, error) {
 	return &p, nil
 }
 
+func ValidatePoolService(ctx context.Context, pool string, services []string) error {
+	poolServices, err := servicemanager.Pool.Services(ctx, pool)
+	if err != nil {
+		return err
+	}
+	for _, svc := range services {
```
arthurcgc, line 613: this should be a function, i.e:  `contains(service, services) bool {...}`

service/actions.go
```diff
@@ -214,7 +215,7 @@ var notifyUpdateServiceInstance = action.Action{
 	MinParams: 4,
 }
 
-type bindPipelineArgs struct {
+type bindAppPipelineArgs struct {
```
arthurcgc, line 218: I think this could also be a generic struct, instead of bindAppPipelineArgs we could do something like:

```
in: binder.go we could add all the other methods necessary for bind.App interface, this is just so you get the idea
type BindableObj interface {
	GetName() string
	GetPool() string
}

type bindAppPipelineArgs struct {
	bindableObject             bind.BindableObj
	writer          io.Writer
	serviceInstance *ServiceInstance
	appParams          BindAppParameters
	event           *event.Event
	requestID       string
	shouldRestart   bool
	forceRemove     bool
}
```

service/service_instance.go
```diff
@@ -204,6 +207,17 @@ func (si *ServiceInstance) FindApp(appName string) int {
 	return index
 }
 
+func (si *ServiceInstance) FindJob(jobName string) int {
+	index := -1
+	for i, name := range si.Jobs {
+		if name == jobName {
+			index = i
```
arthurcgc, line 214:
```suggestion
			return i
```
😝

service/service_instance.go
```diff
@@ -204,6 +207,17 @@ func (si *ServiceInstance) FindApp(appName string) int {
...
+			break
+		}
+	}
+	return index
```
arthurcgc, line 218:
```suggestion
	return -1
```

arthurcgc (changes requested):

PR #2605, opened 2023-03-21
feat: kubernetes provisioner event tracking for jobs

arthurcgc:


event/event.go
```diff
@@ -95,6 +95,7 @@ var (
 	TargetTypeGlobal          = TargetType("global")
 	TargetTypeApp             = TargetType("app")
 	TargetTypeJob             = TargetType("job")
+	TargetTypeCronJob         = TargetType("cronjob")
```
wpjunior, line 98: I suggest to not have other kind, please use CustomData to differ job from cronjob

arthurcgc: I disagree, this is just to group the events by it's appropriate "owner" type, the user will want to list the events by the CronJob name and if we use the "Job" directive for everything it will create ambiguity among users.
Furthermore, both types currently exist in this current tsuru build, so it makes perfect sense to differentiate from them in `job list`.

wpjunior: semantically we have just a resource job in restapi, another name could be confuses the user.

provision/kubernetes/job_test.go
```diff
@@ -99,6 +100,7 @@ func (s *S) TestProvisionerCreateCronJob(c *check.C) {
 										"tsuru.io/job-name":            "myjob",
 										"tsuru.io/job-pool":            "test-default",
 										"tsuru.io/job-team":            "admin",
+										"tsuru.io/tsuru-job":           "true",
```
wpjunior, line 103: following the boolean semantics, could be `tsuru.io/is-tsuru-job`

provision/labels.go
```diff
@@ -37,6 +37,7 @@ const (
 	LabelAppVersion   = "app-version"
 	LabelAppTeamOwner = "app-team"
 
+	LabelIsTsuruJob   = "is-tsuru-job"
```
nettoclaudio, line 40: Why not just `is-job`?

arthurcgc: Guess you're right... Lazy thinking on my part, I just thought we could kill two labels with one using this, but makes no sense since I wasn't brave enough to remove `is-tsuru` label 😜

provision/kubernetes/monitor.go
```diff
@@ -443,11 +551,54 @@ var InformerFactory = func(client *ClusterClient, tweak internalinterfaces.Tweak
...
+	args := newInformerFactoryArgs(client, tsuruServiceTweakFunc())
 	cli, err := VPAClientForConfig(args.restConfig)
 	if err != nil {
 		return nil, err
 	}
 	tweak := vpaInternalInterfaces.TweakListOptionsFunc(args.tweak)
 	return vpaInformers.NewFilteredSharedInformerFactory(cli, args.resync, metav1.NamespaceAll, tweak), nil
 }
+
+func createJobEvent(job *batchv1.Job, evt *apiv1.Event) {
```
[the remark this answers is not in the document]
arthurcgc: agreed

PR #2610, opened 2023-04-06
Cronjobsv1 logs

arthurcgc:


api/job.go
```diff
@@ -584,6 +587,69 @@ func unbindJobServiceInstance(w http.ResponseWriter, r *http.Request, t auth.Tok
...
+	urlValues := r.URL.Query()
+	follow, _ := strconv.ParseBool(urlValues.Get("follow"))
+	jobName := urlValues.Get(":job")
+	j, err := getJob(ctx, jobName)
+	if err != nil {
+		return err
+	}
+	allowed := permission.Check(t, permission.PermJobRead,
+		contextsForJob(j)...,
+	)
```
nettoclaudio, line 623: We need a specific permission to read logs from jobs.

types/log/log.go
```diff
@@ -0,0 +1,58 @@
...
+type LogService interface {
+	Enqueue(entry *LogEntry) error
+	Add(name, tsuruType string, message, source, unit string) error
+	List(ctx context.Context, args ListLogArgs) ([]LogEntry, error)
+	Watch(ctx context.Context, args ListLogArgs) (LogWatcher, error)
+}
+
+type LogServiceProvision interface {
+	Provision(name, tsuruType string) error
+	CleanUp(name, tsuruType string) error
```
nettoclaudio, line 37:
```suggestion
	Provision(name string, type LogType) error
	CleanUp(name string, type LogType) error
```

types/log/log.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+type LogWatcher interface {
+	Chan() <-chan LogEntry
+	Close()
+}
+
+type LogService interface {
+	Enqueue(entry *LogEntry) error
+	Add(name, tsuruType string, message, source, unit string) error
```
nettoclaudio, line 30:
```suggestion
	Add(name string, type LogType, message, source, unit string) error
```

types/log/log.go
```diff
@@ -0,0 +1,58 @@
...
+	Type         LogType
+	Source       string
+	Units        []string
+	Limit        int
+	InvertSource bool
+	Token        auth.Token
+}
+
+type LogEntry struct {
+	MongoID bson.ObjectId `bson:"_id,omitempty" json:"-"`
```
nettoclaudio, line 51: Do we yet store the log entries in MongoDB? I really don't think so... Can check again? If we don't please remove this field.

types/app/log.go
```diff
@@ -44,6 +46,7 @@ type Applog struct {
 	Date    time.Time
 	Message string
 	Source  string
-	AppName string
+	Name    string
```
nettoclaudio, line 49: Are you sure this change don't break the client side?

nettoclaudio (changes requested):

api/log_test.go
```diff
@@ -52,11 +52,11 @@ func (s *S) TestAddLogsHandler(c *check.C) {
 	c.Assert(err, check.IsNil)
 	baseTime = baseTime.Local()
 	bodyStr := `
-	{"date": "2015-06-16T15:00:00.000Z", "message": "msg1", "source": "web", "appname": "myapp1", "unit": "unit1"}
```
wpjunior, line 55: this change may break the tsuru client ? @arthurcgc

arthurcgc: ?

arthurcgc: I didn't understand the relation of what you said with the line of code you just highlighted... Are you asking the same thing @nettoclaudio asked? If so, it doesn't break the client....

wpjunior: Ok, let me explain, you changed the contract of log api renaming field appname to name, but, fortunately, tsuru-client does not care about with this field:  https://github.com/tsuru/tsuru-client/blob/master/tsuru/client/log.go#L107

nettoclaudio (approved):
lgtm

PR #2611, opened 2023-04-11
Drop bind of units on services

wpjunior:
Looking up of all implementations of service this feature become no-op after the arise of kubernetes, after that, the tsuru api will be more like a control plane api, avoiding a dependency of api to bind a pod on a service.


nettoclaudio:
Honestly, I think there's yet a room for this feature. Although we don't we use it internally anymore, OSS services might take advantage of it to do their business logic. Additionally, this change breaks the Tsuru Service API definition, are we sure we want to do it? I would like hear more opnions about it. :thought_balloon:

wpjunior:
Honestly, I think whether user wants to bind podIP on some service is more accurate to plug directly into k8s cluster, this MR turns our API more slim, then I no see problems to deprecate this feature in favor of less data-plane execution on API.

nettoclaudio:
> user wants to bind podIP on some service is more accurate to plug directly into k8s cluster
What about Tsuru Services that are not running in the app's cluster? Do you think it's OK drop completely support for them?

PR #2612, opened 2023-04-11
fix(api/app): preventing to add non-UNIX env var names

nettoclaudio:
This PR prevents user adding env var that aren't UNIX compliant onto app.


wpjunior:
ok LGTM

wpjunior (approved):

crgisch (commented):
LGTM

PR #2613, opened 2023-04-12
Add support to envvars with jobs

wpjunior:


crgisch (approved):
LGTM

nettoclaudio:
So, I think we should follow the pattern used on app resource to set/unset env vars:
* POST on `/jobs/:name/env` to set them
* DELETE on `/jobs/:name/env` to unset them

wpjunior:
@nettoclaudio do you think to also reuse same CLI commands ?

nettoclaudio:
I think so. This way we can ensure fine-grained permissions (e.g. creating an user that can only change set/unset env vars), keep the pattern of app's commands (since users are already familiar with), keep the pattern of Terraform resources and so on.

api/app.go
```diff
@@ -1200,16 +1200,25 @@ func setEnv(w http.ResponseWriter, r *http.Request, t auth.Token) (err error) {
 }
 
 func IsEnvVarValid(envs ...apiTypes.Env) error {
```
arthurcgc, line 1202: I would change the name to: ValidateApiEnvVars

arthurcgc, line 1202: why not just ? `envs []apiTypes.Env`

api/app.go
```diff
@@ -1200,16 +1200,25 @@ func setEnv(w http.ResponseWriter, r *http.Request, t auth.Token) (err error) {
...
+	return isEnvsNamesVarValid(envNames...)
+}
+
+func isEnvsNamesVarValid(envs ...string) error {
 	var errs errors.MultiError
 
 	for _, e := range envs {
-		if isInternalEnv(e.Name) {
-			errs.Add(fmt.Errorf("cannot change an internal environment variable (%s): %w", e.Name, apiTypes.ErrWriteProtectedEnvVar))
+		if isInternalEnv(e) {
```
arthurcgc, line 1215: I would extract it to a function
```suggestion
func isValid(env string) error {
	if isInternalEnv(env) {
		return fmt.Errorf("cannot change an internal environment variable (%s): %w", env, apiTypes.ErrWriteProtectedEnvVar)
	}

	if err := isEnvVarName(env); err != nil {
		return fmt.Errorf("%q is not valid environment variable name: %w", env, err)
	}

	return nil
}
```

arthurcgc: I would also change the name isEnvVarName to isUnixLike

nettoclaudio (approved):
lgtm

PR #2615, opened 2023-04-24
Use job labels following the standard: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

wpjunior:


crgisch (approved):
LGTM

PR #2620, opened 2023-06-01
Suspended Jobs

arthurcgc:
* Enable setting the "Manual" attribute to a Cronjob.
* Every job is now necessarily a Cronjob, manual jobs are created with the cronjob.spec.suspended = true, label tsuru.io/job-manual = true and cronjob.spec.schedule = "* * 31 2 *"
* Every job creation now requires a unique job name to be provided


job/job.go
```diff
@@ -100,9 +100,9 @@ func (*jobService) DeleteFromProvisioner(ctx context.Context, job *jobTypes.Job)
 //
 //  1. Save the job in the database
 //  2. Provision the job using the provisioner
-func (*jobService) CreateJob(ctx context.Context, job *jobTypes.Job, user *authTypes.User, trigger bool) error {
+func (*jobService) CreateJob(ctx context.Context, job *jobTypes.Job, user *authTypes.User) error {
 	jobCreationErr := jobTypes.JobCreationError{Job: job.Name}
-	if err := buildName(ctx, job); err != nil {
+	if err := validateName(ctx, job); err != nil {
 		jobCreationErr.Err = err
```
wpjunior, line 106: @arthurcgc could we validate some errors:

* invalid schedule syntax
* missing required fields such as schedule

arthurcgc: @wpjunior we already validate schedule, even though k8s tries to do that too, right here: 
* [schedule validation](https://github.com/tsuru/tsuru/blob/ab41e8c6ac04f7594a1f45a3cd861b1f680ab26e/job/job.go#L390)
* [job validation](https://github.com/tsuru/tsuru/blob/ab41e8c6ac04f7594a1f45a3cd861b1f680ab26e/job/job.go#L115)

wpjunior: nice!

wpjunior (approved):

PR #2621, opened 2023-06-05
Drop Router V1

wpjunior:


app/app.go
```diff
@@ -842,10 +846,13 @@ func (app *App) AddUnits(n uint, process, versionStr string, w io.Writer) error
 		&reserveUnitsToAdd,
 		&provisionAddUnits,
 	).Execute(app.ctx, app, n, w, process, version)
-	rebuild.RoutesRebuildOrEnqueueWithProgress(app.Name, w)
```
wpjunior, line 845: Maybe we can remove this call

PR #2633, opened 2023-07-19
Enforce consistent yaml formatting

tcarreira:
yaml files in this project are inconsistent formatted. This fixes:
- quoted fields with `"`
- array fields with no indent

This PR aims to fix that, ensuring it will stay well formatted on the CI stage.

note: there are 2 commits: the first is my changes, the second is running only the formatter


nettoclaudio (approved):
lgtm

PR #2634, opened 2023-07-20
feat: Enable Service Account Job Annotations

arthurcgc:
Also: DestroyJob now deletes the job's SA

### DIFF
--- a/provision/kubernetes/job.go
+++ b/provision/kubernetes/job.go
@@ -70,6 +70,7 @@ func buildJobSpec(job *jobTypes.Job, client *ClusterClient, labels, annotations 
 						Env:       envs,
 					},
 				},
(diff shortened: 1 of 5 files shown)


provision/kubernetes/job.go
```diff
@@ -223,6 +224,9 @@ func (p *kubernetesProvisioner) DestroyJob(ctx context.Context, job *jobTypes.Jo
 		return err
 	}
 	namespace := client.PoolNamespace(job.Pool)
+	if err := client.CoreV1().ServiceAccounts(namespace).Delete(ctx, serviceAccountNameForJob(*job), metav1.DeleteOptions{}); err != nil {
+		return err
```
wpjunior, line 228: @arthurcgc ignore when it is not found

wpjunior (approved):

PR #2650, opened 2023-09-11
Get default cluster configs from tsuruconf

morpheu:
PR to check tsuruconf for default cluster configs when not explicit set


provision/kubernetes/cluster.go
```diff
@@ -660,7 +609,14 @@ func (c *ClusterClient) configForContext(context, key string) string {
 	if v, ok := c.CustomData[context+":"+key]; ok {
 		return v
 	}
-	return c.CustomData[key]
+	if v, ok := c.CustomData[key]; ok {
+		return v
+	}
+	data, err := config.GetString("clusters:defaults:" + key)
```
wpjunior, line 615: @morpheu have risky to raise an exception of key not found here?

morpheu: In such case always returns an empty string. Most of methods were check also for empty CustomData, so ignoring all config errors and treat them as empty string seems ok to me

wpjunior (approved):

PR #2652, opened 2023-09-19
Job cleanup

arthurcgc:
Add default parameters to activeDeadlineSeconds and TTLSecondsAfterFinished


job/job.go
```diff
@@ -442,6 +445,13 @@ func validateTeamOwner(ctx context.Context, job *jobTypes.Job, p *pool.Pool) err
 	return &tsuruErrors.ValidationError{Message: msg}
 }
 
+func buildActiveDeadline(ctx context.Context, j *jobTypes.Job) {
+	defaultActiveDeadline := int64(86400)
```
wpjunior, line 449: @arthurcgc we will fill the defaultActiveDeadline just on creation ?

PR #2664, opened 2023-10-26
provision/kubernetes: use k8s jobs instead pods for units

morpheu:


provision/kubernetes/job.go
```diff
@@ -234,44 +235,51 @@ func (p *kubernetesProvisioner) DestroyJob(ctx context.Context, job *jobTypes.Jo
...
+		for _, pod := range pods {
+			restarts = containersRestarts(pod.Status.ContainerStatuses)
+			if *restarts > 0 {
+				break
+			}
+		}
+		switch {
+		case k8sJob.Status.Failed > 0:
+			status = provision.StatusError
+		case k8sJob.Status.Succeeded > 0:
```
wpjunior, line 271: there is a case that shows status error:
example:
failed=1
succeeded=1
my my opnion should show a succeeded response, what do you think?

provision/kubernetes/job.go
```diff
@@ -234,44 +235,51 @@ func (p *kubernetesProvisioner) DestroyJob(ctx context.Context, job *jobTypes.Jo
...
 
-		createdAt := pod.CreationTimestamp.Time.In(time.UTC)
+		createdAt := k8sJob.CreationTimestamp.Time.In(time.UTC)
 		units = append(units, provision.Unit{
-			ID:        pod.Name,
-			Name:      pod.Name,
-			IP:        pod.Status.HostIP,
+			ID:        k8sJob.Name,
+			Name:      k8sJob.Name,
 			Status:    status,
```
wpjunior, line 281: @morpheu what do you think about show IP of running pod?

morpheu: I omitted intentionally think on future scenarios where a job could spawn multiple pods

provision/kubernetes/job.go
morpheu: On a use case where 1 job spawn 1 single pod this scenario will never occurs. Considering a multi pod scenario, I really don't know which is the best approach to show - maybe change job return showing how many pods were successfully completed and how many don't

provision/kubernetes/job.go
wpjunior: Yeah! in case of multiple pods unique IP lose the sense

provision/kubernetes/job.go
```diff
@@ -215,11 +212,15 @@ func (p *kubernetesProvisioner) JobUnits(ctx context.Context, job *jobTypes.Job)
 	if err != nil {
 		return nil, err
 	}
-	pods, err := p.podsForJobs(ctx, client, []*jobTypes.Job{job})
+	labelSelector := metav1.LabelSelector{MatchLabels: map[string]string{tsuruLabelJobName: job.Name}}
+	listOptions := metav1.ListOptions{
```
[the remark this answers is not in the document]
arthurcgc: actually, provision/labels.go/JobLabels() might be even better

provision/kubernetes/job.go
```diff
@@ -234,44 +235,51 @@ func (p *kubernetesProvisioner) DestroyJob(ctx context.Context, job *jobTypes.Jo
...
-		} else {
-			status = stateMap[pod.Status.Phase]
+		var restarts *int32
+		pods, err := p.getPodsForJob(ctx, client, &k8sJob)
+		if err != nil {
+			return nil, err
+		}
+		for _, pod := range pods {
+			restarts = containersRestarts(pod.Status.ContainerStatuses)
+			if *restarts > 0 {
```
arthurcgc, line 264: I don't quite get this logic, if we find a pod with *restarts > 0 why do we break and stop counting them?
Maybe I'm missing something, but shouldn't we count every pod related to the Job?

wpjunior (approved):

PR #2670, opened 2023-11-23
tag validator service

wpjunior:


api/server.go
```diff
@@ -191,6 +192,10 @@ func setupServices() error {
 	if err != nil {
 		return errors.Wrapf(err, "could not initialize job service")
 	}
+	servicemanager.Tag, err = tag.TagService()
```
arthurcgc, line 195: I think this could be an OptIn from the command line, since tag validation services are **_unusual_**... 
Something like: `./api --tag-validation` WDYT?

PR #2672, opened 2023-11-30
Tsuru event TTL

arthurcgc:


crgisch (approved):
LGTM 👍

provision/kubernetes/job_test.go
```diff
@@ -684,7 +685,9 @@ func (s *S) TestCreateJobEvent(c *check.C) {
 					Reason:  "SuccessfulCreate",
 					Type:    "Normal",
 				}
-				createJobEventAndMetrics(j, evt)
+				wg := &sync.WaitGroup{}
+				wg.Add(1)
+				createJobEvent(j, evt, wg)
```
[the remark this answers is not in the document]
arthurcgc: ?? 
Actually no.. I didn't spawn a goroutine, there's no need to wg.Wait()... The only reason I added wg.Add(1) is because we decrement it at the defer statement

Dockerfile
```diff
@@ -16,3 +16,6 @@ EXPOSE 8080
 ENTRYPOINT ["/usr/local/bin/tsurud"]
 CMD ["api"]
 COPY --from=builder /go/src/github.com/tsuru/tsuru/build/tsurud /usr/local/bin/tsurud
+COPY gke-auth-plugin_Linux_x86_64.tar.gz /tmp
```
wpjunior, line 19: @arthurcgc please revert this lines

provision/kubernetes/job.go
```diff
@@ -29,6 +31,7 @@ import (
 const (
 	promNamespace = "tsuru"
 	promSubsystem = "job"
+	expireTTL     = time.Hour * 24 // 1 day
```
wpjunior, line 34: @arthurcgc I prefer to have two variables to cause curiosity of code reader to understand logic bellow.

```
frequentExpireTTL = time.Hour * 24
occasionalExpireTTL = time.Hour
```

provision/kubernetes/monitor.go
```diff
@@ -167,7 +167,11 @@ func (c *clusterController) startJobInformer() error {
 			if job.Labels["tsuru.io/is-tsuru"] != "true" {
 				return
 			}
-			createJobEventAndMetrics(job, evt)
+			wg := &sync.WaitGroup{}
```
wpjunior, line 170: increment metrics on prometheus is so fast, you don't need to spawn a goroutine for this, you could remove waitGroup
@arthurcgc

arthurcgc: It's not about how fast it goes, it's idiomatic and clean. Your proposal seems more like a preference than a recommendation.

wpjunior: Yeah, I prefer to spawn a goroutine when is favored, to avoid stimulus to other developers to use goroutine everywhere.

but it is a philosophical approach, your case is not harmful, cause is a small context of use of goroutine.

PR #2676, opened 2024-01-10
docs: Fix typo in Installing Tsuru in a local Kubernetes

gvicentin:
When following the doc **Installing Tsuru in a local Kubernetes cluster with Helm** there is an extra `s` in [Configuring Tsuru](https://docs.tsuru.io/main/installing/installing-minikube.html#configuring-tsuru) that will make `tsuru cli` not find the API.

In order to generate the docs for validation I had to do the following:
1. Create a virtual environment with Python 2.7
2. Pin a couple of Python's dependencies so Sphinx can start properly.
3. Move `releases RELEASES` to the try block for running locally without problems.
4. Fix some old reference targets in the docs


arthurcgc (approved):
lgtm! tks @gvicentin

PR #2686, opened 2024-03-25
Drop tsuru_unit_agent unused code

wpjunior:


api/app.go
```diff
@@ -1233,7 +1233,7 @@ func isInternalEnv(envKey string) bool {
 }
 
 func internalEnvs() []string {
-	return []string{"TSURU_APPNAME", "TSURU_APP_TOKEN", "TSURU_SERVICE", "TSURU_APPDIR"}
```
wpjunior, line 1236: @morpheu I think that TSURU_SERVICE was never used, this is a kind of typo, right?

PR #2687, opened 2024-03-26
Add validation to avoid conflict with service envs

wpjunior:


crgisch (approved):
LGTM in relation to APPs
We might need the same for jobs

PR #2700, opened 2024-06-14
Migrate App to types part 2

wpjunior:


api/shell.go
```diff
@@ -6,6 +6,7 @@ package api
 
 import (
 	"bytes"
+	stdContext "context"
```
mauri870, line 9: I'd suggest leaving this as context and using `tsurucontext` or similar for the tsuru one. Most developers will imediatelly think of context when they read `context.Something`.

mauri870: I just noticed the `stdcontext` approach it is used in most places that import both contexts. In this case consistency should be the priority.

PR #2721, opened 2024-08-18
Fixing and improving local development tools

gvicentin:
Tested on Mac and Linux with the following specs:

```
Darwin 23.5.0 Darwin Kernel Version 23.5.0: Wed May  1 20:14:38 PDT 2024; root:xnu-10063.121.3~5/RELEASE_ARM64_T6020 arm64
```

```
Linux  6.8.0-40-generic #40-Ubuntu SMP PREEMPT_DYNAMIC Fri Jul  5 10:34:03 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux
```

## Fix and improvements:

1. Compose file:
    1. Added a volume for mongodb data. Now we don’t loose data between restarts. If needed, we can still reset the database using make `local.cleandb`. 
    2. Tsuru API now runs on a separated profile. This is useful because most of the time we will compile and run the api directly. We can still spin up the tsuru-api container using `docker compose --profile tsuru-api up -d`.
2. Created a shell cli for centralizing local development utilities: `./misc/local-dev.sh`.
3. Now we check if minikube is running before starting it. This helps to speed up things since minikube start is not idempotent. 
4. Check if `socker_vmnet` service is running on MacOS before starting minikube cluster.
5. Now we check if the loopback interface has already been aliased with fake ip before assigning it (it wouldn’t work with command `ip`)
6. Refactored and simplified Makefile targets for local dev.

## What this changes:

With this, the development setup will be much simpler:

Setup environment for the first time:

```bash
make local.setup
```

1. This will configure the loopback interface if needed.
2. Generate tsuru config files.
3. Start mikube if needed. It will consider if we are running on Mac and take the appropriated actions.
4. Spin up dependencies containers with docker compose.
5. Create tsuru root user if doesn’t exist already.
6. Configure default cluster if doens’t exist already.
7. Add local-dev tsuru target.
8. Stop tsuru-api container.

Then we can use the following 

```
make local.run
```
1. Check if setup was completed.
2. Start dependencies if needed
3. Build and start Tsuru API

## We also have the opposite commands:

For stopping local dependencies

```bash
make local.stop
```

Cleanup local environment for good:

```bash
make local.cleanup
```


gvicentin:
Hey @crgisch and @wpjunior.

I was considering adding something like this to the README.md. Let me know if you guys think there's a better place for this little doc.

## Local development

### Dependencies

Before starting, make sure you have the following tools installed:

* [docker](https://docs.docker.com/engine/install) (or [podman](https://podman.io/docs/installation))
* [minikube](https://minikube.sigs.k8s.io/docs/start)
* [go](https://go.dev/dl/)
* [yq](https://github.com/mikefarah/yq#install)

Also we need to have [tsuru-client](https://docs.tsuru.io/stable/using/install-client.html) installed for interacting with the API. Go ahead and install it if you haven't already.

For MacOS users, we recommend using the **_qemu_** driver with **_socket_vmnet_** for **_minikube_** clusters. For more information about installing **qemu** and **socket_vmnet**, check the following links:

* [qemu](https://www.qemu.org/download/)
* [socket_vmnet](https://github.com/lima-vm/socket_vmnet)

**Note**: If you're using alternatives that are compatible with Docker, such as Podman, make sure specify the DOCKER variable with the correct binary when running make commands, e.g. make local.run DOCKER=podman.

### Running local environment

To run Tsuru API locally, we need to first setup the local environment . This will create the default configuration and make sure we have everything needed to start playing around.

```bash
make local.setup 
```

After the setup, we don't need to run the previous command again. From now on, we can just use the following command to start dependencies and run the api:

```bash
make local.run
```

After the Tsuru API is up and running, we can use a new terminal to setup our CLI to point to the local-dev target using the command bellow (Tsuru's targets work similarly to kubectl config context):

```bash
tsuru target set local-dev
```

To validate that everything is working as expected, you can list the clusters managed by your Tsuru API instance. You should see your local minikube cluster as the default provisioner:

```bash
tsuru cluster list
```

### Cleaning up

For stopping the dependencies:

```bash
make local.stop
```

If you want to reset your environment or remove Tsuru dependencies for good:

```bash
make local.cleanup
```

misc/local-dev.sh
```diff
@@ -0,0 +1,268 @@
...
+}
+
+exec_setup_tsuru_target() {
+    local target_host=${1:-"$FAKE_HOST_IP"}
+    local target_port=${2:-"$TSURU_API_PORT"}
+    local target_name=${3:-"local-dev"}
+
+    # Add the Tsuru target, ignore the output and errors because the target may already exist
+    echo "Setting up the Tsuru target $target_name at http://${target_host}:${target_port}..."
+    tsuru target add "$target_name" "http://${target_host}:${target_port}" &> /dev/null || true
```
crgisch, line 162: Something we talked about yesterday would be to add `--set-current` just as an extra safety measure

crgisch (approved):
Nicely done!
This is awesome 💯

misc/local-dev.sh
crgisch: Actually, breaking it in two commands would be better:
Leave the `tsuru target` as it is now
Add `tsuru target set "$target_name"` (so if the target already exists and the target adds fails, we still switch to development target after that)

PR #2727, opened 2024-09-02
fix: preserve main collection when running make test

gvicentin:
There is also a fix for a **oidc_test** where a new user is inserted in the main collection as well.

### DIFF
--- a/auth/oidc/oidc_test.go
+++ b/auth/oidc/oidc_test.go
@@ -19,6 +19,7 @@ import (
 	"github.com/lestrrat-go/jwx/v2/jwk"
 	"github.com/tsuru/config"
 	"github.com/tsuru/tsuru/auth"
+	"github.com/tsuru/tsuru/db/storagev2"
 	check "gopkg.in/check.v1"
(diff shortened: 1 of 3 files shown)


auth/oidc/oidc_test.go
```diff
@@ -41,6 +41,9 @@ func (s *AuthSuite) SetUpSuite(c *check.C) {
 		json.NewEncoder(w).Encode(s.jwkKeySet)
 	}))
 
+	config.Set("database:url", "127.0.0.1:27017?maxPoolSize=100")
+	config.Set("database:name", "oidc_tests")
+
```
wpjunior, line 46: please call `storagev2.Reset()` to ensure that old singleton database client is destructed

wpjunior:
Great job @gvicentin 👏

auth/oidc/oidc_test.go
gvicentin: added, thanks 🥇

wpjunior (approved):

PR #2730, opened 2024-09-11
feat: list services instances by tags

prbn97:


api/service.go
```diff
@@ -56,10 +56,14 @@ func serviceList(w http.ResponseWriter, r *http.Request, t auth.Token) error {
 	if err != nil {
 		return err
 	}
+
+	tags := r.URL.Query()["tag"]
+	sInstancesFilteredByTags := filterServicesInstancesByTags(sInstances, tags)
```
wpjunior, line 61: @prbn97 you could reuse the same variable

```suggest
sInstances := filterServicesInstancesByTags(sInstances, tags)
```

api/service.go
```diff
@@ -505,3 +509,25 @@ func contextsForServiceProvision(s *service.Service) []permTypes.PermissionConte
...
+	if len(tags) == 0 {
+		return sInstances
+	}
+
+	tagMap := make(map[string]struct{}, len(tags))
+	for _, tag := range tags {
+		tagMap[tag] = struct{}{}
+	}
+
+	var filteredInstances []service.ServiceInstance
```
wpjunior, line 523: for performace reasons you may initialize a slice with high capacity like this:
```suggest
filteredInstances := make([]service.ServiceInstance, 0, len(sInstances))
```

wpjunior (commented):
Great work @prbn97, welcome aboard! 🚀 🚀

api/service.go
prbn97: excellent tip fo perfomace, thanks!

api/service.go
```diff
@@ -505,3 +509,25 @@ func contextsForServiceProvision(s *service.Service) []permTypes.PermissionConte
...
+	tagMap := make(map[string]struct{}, len(tags))
+	for _, tag := range tags {
+		tagMap[tag] = struct{}{}
+	}
+
+	filteredInstances := make([]service.ServiceInstance, 0, len(sInstances))
+	for _, service := range sInstances {
+		for _, sTag := range service.Tags {
+			if _, exists := tagMap[sTag]; exists {
+				filteredInstances = append(filteredInstances, service)
```
arthurcgc, line 527: nifty little hack to simplify the code:
```suggestion
				return append(filteredInstances, service)
```
you can then get rid of "break" and return nil (or an empty slice) at the end which makes it clearer 😄

wpjunior: @arthurcgc nice point, but I think the desired behavior is to remove the break word, cause the filter need to check the full array

wpjunior:
@prbn97 I think that you could easy achieve modifying the GetServiceInstancesByServices function to have similar behavior that you had implemented

wpjunior (approved):

gvicentin (approved):

PR #2734, opened 2024-10-24
feat: customize auto scale down behavior

infezek:
This PR aims to customize the auto scale down behavior, allowing the user to configure it according to their needs or accept the default values from Tsuru


types/provision/autoscale.go
```diff
@@ -15,6 +15,17 @@ type AutoScaleSpec struct {
...
 	Version    int                   `json:"version"`
+	Behavior   BehaviorAutoScaleSpec `json:"behavior,omitempty"`
+}
+
+type BehaviorAutoScaleSpec struct {
+	ScaleDown *ScaleDownPoliciy `json:"scaleDown"`
+}
+
+type ScaleDownPoliciy struct {
+	StabilizationWindow   *int32 `json:"stabilizationWindow"`
```
wpjunior, line 26: omitempty sound good here for everyfield @infezek

wpjunior (approved):

PR #2740, opened 2024-12-26
Feat/shell debug

infezek:
This feature adds an ephemeral container, 
Aiming to enable the use of debugging tools for applications, even with Alpine-based images


ravilock:
LGTM

Issue #2574, opened 2022-10-07
Support multiple healthchecks for Tsuru app


nettoclaudio:
**Is your feature request related to a problem? Please describe.**
Tsuru only supports a single healthcheck configuration for the whole app. For apps with multiple processes only the "web" process can configure healthcheck - as you can see on [k8s](https://github.com/tsuru/tsuru/blob/main/provision/kubernetes/deploy.go#L657-L663) and [docker](https://github.com/tsuru/tsuru/blob/main/provision/docker/actions.go#L389-L394) provisioners.

**Describe the solution you'd like**
In order to avoid breaking changes in the Tsuru config file for apps (aka [`tsuru.yaml`](https://docs.tsuru.io/stable/using/tsuru.yaml.html)), the safest approach would be to create a new array property to hold the multiple health check configs. Each array item must provide the probe configs along the process name (which is required).

```yaml
healthchecks:
- process: web
  path: /healthz
...
- process: worker
  command: ["/path/to/my/script.sh", "--healthcheck"]
```

But I also like the idea of supporting both solutions in the existing `healthcheck` property. So according to the type of `healthcheck` property, Tsuru does a thing or something else. When an array is presented, multiple probe configs are supported, otherwise, the legacy behavior is kept.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #2741, opened 2025-01-06
Feat - Multiple App Healthchecks

ravilock:
The ideia is to introduce a new format to tsuru.yaml. It might deprecate Procfile usage.

```yaml
processes:
  - name: web
    command: python app.py
    healthcheck:
      path: /
      scheme: http
  - name: web-secondary
    command: python app2.py
    healthcheck:
      path: /
      scheme: http
```

### DIFF
--- a/provision/kubernetes/deploy.go
+++ b/provision/kubernetes/deploy.go
@@ -501,7 +501,8 @@ func createAppDeployment(ctx context.Context, client *ClusterClient, depName str
 	var hcData hcResult
 	// NOTE: Here is the code that create probes for HEALTHCHECK!
 	if len(yamlData.Processes) > 0 {
-		healthcheck, err := yamlData.GetHCFromProcessName(process)
+		var healthcheck *provTypes.TsuruYamlHealthcheck
+		healthcheck, err = yamlData.GetHCFromProcessName(process)
 		if err != nil {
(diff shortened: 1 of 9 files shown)


ravilock:
closes #2574

provision/kubernetes/deploy.go
```diff
@@ -507,8 +499,18 @@ func createAppDeployment(ctx context.Context, client *ClusterClient, depName str
 		return false, nil, nil, errors.WithStack(err)
 	}
 	var hcData hcResult
-	if process == webProcessName && len(processPorts) > 0 {
-		//TODO: add support to multiple HCs
+	// NOTE: Here is the code that create probes for HEALTHCHECK!
+	if len(yamlData.Processes) > 0 {
+		var healthcheck *provTypes.TsuruYamlHealthcheck
+		healthcheck, err = yamlData.GetHCFromProcessName(process)
```
morpheu, line 505: If user defines a usual global healthcheck block outside the new process structure it could be used instead returning an error ? Or can we send some kind of warning informing it will be ignored ?

ravilock: We could return a warning stating that global healthchecks and `Procfile` are going to be ignored when using the new process structure.

morpheu (approved):

PR #2743, opened 2025-01-29
registry: tweaks for image cleanup

morpheu:
- using docker config style and dropping old registry credential config for all clusters
- allowing using separated credentials for different clusters instead single auth basic credential
- fix on RemoveAppImages when app is removed
- add support for docker registry realm style authentication

### DIFF
--- a/registry/registry.go
+++ b/registry/registry.go
@@ -202,6 +202,9 @@ request:
 				continue
 			}
 			if resp.StatusCode == http.StatusUnauthorized && resp.Header.Get("WWW-Authenticate") != "" && r.checkTokenIsValidForRenew() {
+				if resp.Body != nil {
+					resp.Body.Close()
+				}
 				to := auth.TokenOptions{}
 				to.Username = r.authConfig.Username
 				to.Secret = r.authConfig.Password
--- a/registry/registry.go
+++ b/registry/registry.go
@@ -66,6 +66,9 @@ func RemoveImage(ctx context.Context, imageName string) error {
 		return errors.New("invalid empty image name")
 	}
 	registry, image, tag := image.ParseImageParts(imageName)
+	if registry == "" {
+		return errors.New("invalid empty registry")
+	}
 	r := &dockerRegistry{registry: registry}
 	err := r.registryAuth(ctx, imageName)
 	if err != nil {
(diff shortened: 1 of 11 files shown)


registry/registry.go
```diff
@@ -205,51 +196,171 @@ func (r *dockerRegistry) doRequest(ctx context.Context, method, path string, hea
...
+	if r.token == "" || (r.token != "" && time.Now().After(r.expires)) {
+		return true
+	}
+	return false
+}
+
+func (r *dockerRegistry) registryAuth(ctx context.Context, image string) error {
+	var err error
+	var config *configfile.ConfigFile
+	if r.cluster != nil {
```
wpjunior, line 261: @morpheu where r.cluster become filled?

registry/registry.go
```diff
@@ -205,51 +196,171 @@ func (r *dockerRegistry) doRequest(ctx context.Context, method, path string, hea
...
+				respAuth.IssuedAt = time.Now()
+			}
+			if respAuth.ExpiresIn == 0 {
+				respAuth.ExpiresIn = defaultExpiration
 			}
+			if exp := respAuth.IssuedAt.Add(time.Duration(float64(respAuth.ExpiresIn)*0.9) * time.Second); time.Now().Before(exp) {
+				r.expires = exp
+			}
+			r.token = respAuth.Token
+			goto request
```
wpjunior, line 231: @morpheu we dont have control to avoid loop of renew, we could implement a max retries

registry/registry.go
morpheu: hummm seems to be a dead code :( - removed it on new commit

registry/registry.go
```diff
@@ -179,77 +168,186 @@ func (r dockerRegistry) removeImagePath(ctx context.Context, path string) error
...
+			}
+			req.Header = http.Header{}
+			for k, v := range headers {
+				req.Header.Set(k, v)
+			}
+			r.fillAuthCredentials(req)
+			resp, err = r.client.Do(req)
 			if _, ok := err.(net.Error); ok {
 				continue
 			}
```
wpjunior, line 203: @morpheu you miss to close the Request with
```
defer req.Body.Close()
```

to simplify that you could extract the inner loop to a function

morpheu: Added missing `resp.Body.Close()` inside fetchAuthToken logic

wpjunior (approved):

PR #2752, opened 2025-02-27
Fix/process command fallback to image

ravilock:
This PR aims to fix an issue where the new `tsuru.yaml` directive `processes` is used, but the `command` of the web process is not defined or is defined as an empty string (`""`).

In those cases, we fallback to use the image command either from `entrypoint` or `cmd`.

### DIFF
--- a/builder/kubernetes/build.go
+++ b/builder/kubernetes/build.go
@@ -390,7 +390,7 @@ func (b *kubernetesBuilder) buildContainerImage(ctx context.Context, app *apptyp
 		}
 
 		// Default to web process name and entrypoint and cmd from container
-		if shouldDefaultToImageCommands(processes) {
+		if len(processes) == 0 {
 			ic := tc.ImageConfig
 			if ic == nil {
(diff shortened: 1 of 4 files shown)


builder/kubernetes/build.go
```diff
@@ -383,13 +383,13 @@ func (b *kubernetesBuilder) buildContainerImage(ctx context.Context, app *apptyp
 		}
 
 		// Default to web process name and entrypoint and cmd from container
-		if len(processes) == 0 {
+		if shouldDefaultToImageCommands(processes) {
```
wpjunior, line 386: @ravilock we may need a else statement to cover some cases

ravilock: Tell me more

wpjunior: @ravilock i think that you could simplify to if len(finalProcesses) == 0, that means that Procfile and tsuru.yaml does not have any command

wpjunior (approved):

PR #2855, opened 2026-04-15
fix(docker): increase mongo ulimit for test stability

ravilock:
Tests rapidly open and close many MongoDB connections, which can exhaust the default file descriptor limit and cause test failures. This sets the `nofile` ulimit to 65536 for the mongo service in docker-compose to prevent such issues during testing.


morpheu:
LGTM

morpheu (approved):

PR #2856, opened 2026-04-15
chore: add script for post integration tests minikube and database cleanup

ravilock:
This script should be safe to run (as it uses the same pattern of INTEGRATION_KUBECONFIG env var).
It does not use tsuru cli (avoiding possible catastrophic issues of deleting actual production apps, jobs, etc)

It is intended to be used after running integration tests.
Integration tests might leave some apps, pools, teams left on tsuru and on cluster state
- Specially when you run `make local.test-ci-integration` and the `TSURU_INTEGRATION_no_rollback` is set to `"true"`
- There are legitimate reasons to disable rollback on integration tests such as validating the state manually after a test fail, so this script comes handy if anything needs to be adjusted after

### DIFF
--- a/scripts/clear-local-integration.sh
+++ b/scripts/clear-local-integration.sh
@@ -1,0 +1,132 @@
+#!/usr/bin/env bash
+
+# Copyright 2025 tsuru authors. All rights reserved.
+# Use of this source code is governed by a BSD-style
+# license that can be found in the LICENSE file.
+set -euo pipefail
+
+readonly KUBECTL=${KUBECTL:-kubectl}
+
+function check_kubectl_config() {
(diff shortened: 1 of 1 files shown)


scripts/clear-local-integration.sh
```diff
@@ -0,0 +1,132 @@
...
+  # remove k8s resources
+  remove_app_services
+  remove_app_deployments
+  remove_app_replicasets
+  remove_app_secrets
+  remove_app_hpas
+  remove_cronjobs
+  remove_jobs
+
+  appImages=$(minikube image ls | grep 5000 | grep app | grep -v "<none>" || true)
```
pauloricardokoch, line 114: what `grep 500` does in this context?

maybe add some comment to explain what this line filters

Issue #2857, opened 2026-04-15
Support per-plan RuntimeClassName for Kubernetes sandboxed runtimes


guilhermebr:
**Is your feature request related to a problem? Please describe.**

There's currently no way to ask Tsuru to run an app's pods under a specific Kubernetes RuntimeClass. Operators who want to offer stronger workload isolation (gVisor, kata-containers, or any other sandboxed runtime handler) have to either patch Tsuru locally or apply cluster-wide runtime defaults that affect every pod.

**Describe the solution you'd like**

Add an optional `RuntimeClassName string` field to `Plan` (and a matching pointer on `PlanOverride` for per-app override). When set, it's threaded into `PodSpec.RuntimeClassName` at deployment time. When empty, behaviour is unchanged — the cluster's default runtime handler is used.

This keeps the change runtime-agnostic: any RuntimeClass registered in the cluster (gvisor, kata-qemu, runsc, wasmedge, etc.) is usable, and Tsuru doesn't need to know about specific runtimes.

**Describe alternatives you've considered**

- **Pool-level config key** (e.g. `runtime-class-name` in cluster CustomData). Could come as a follow-up, but plan-level is the more natural entry point since RuntimeClass choice is usually an app/tier concern rather than a cluster-topology one.
- **Cluster-wide default RuntimeClass in containerd**. Works but is all-or-nothing and can't be overridden per-workload.

**Additional context**

PR with the implementation + tests: #2858. Follows the same shape as `CPUBurst` (optional field on `Plan`, mirror on `PlanOverride`, getter with override fallback, `planOnMongoDB` stays in sync).

PR #2858, opened 2026-04-15
feat(plan): add optional RuntimeClassName field

guilhermebr:
## Summary

Adds an optional `RuntimeClassName` field to `Plan` so apps can be scheduled under a specific Kubernetes RuntimeClass (e.g. `gvisor`, `kata-qemu`). When set, it's threaded into `PodSpec.RuntimeClassName`; when empty, behaviour is unchanged.

Refs #2857.

## Shape

Mirrors the `CPUBurst` precedent:

- `types/app/plan.go`: `Plan.RuntimeClassName string` (`omitempty`) + `PlanOverride.RuntimeClassName *string` + `GetRuntimeClassName()` with override fallback + `MergeOverride` handles empty-string as clear
- `storage/mongodb/plan.go`: matching field on `planOnMongoDB` so the struct conversion stays valid
- `provision/kubernetes/deploy.go`: when `plan.GetRuntimeClassName()` is non-empty, populate `PodSpec.RuntimeClassName`
- API: `ParseInput` on JSON mode already round-trips the field — no handler changes (same as CPUBurst)

## Tests

- `types/app/plan_test.go` (testify): getter + `MergeOverride` semantics
- `provision/kubernetes/deploy_test.go` (gocheck, matching file style): `PodSpec.RuntimeClassName` is set when plan has the field, and remains `nil` (cluster default) when it doesn't

Local: `go test -race` of the affected packages + `golangci-lint v1.64` are clean.

## Intentionally out of scope

- Pool-level override (can be a follow-up if desired)
- CLI flag on `tsuru plan create` — JSON mode works today; happy to add the form-mode flag in a follow-up if that's preferred
- Validation that the RuntimeClass actually exists in the cluster — kubelet admission will fail fast on unknown names

### DIFF
--- a/provision/kubernetes/deploy.go
+++ b/provision/kubernetes/deploy.go
@@ -762,6 +762,11 @@ func createAppDeployment(ctx context.Context, opts *createAppDeploymentOptions) 
 		return false, nil, nil, err
 	}
 
+	var runtimeClassName *string
+	if rcn := plan.GetRuntimeClassName(); rcn != "" {
+		runtimeClassName = &rcn
+	}
+
 	metadata := provision.GetAppMetadata(opts.app, opts.process)
 	podLabels := opts.labels.PodLabels()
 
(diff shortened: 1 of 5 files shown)


wpjunior:
NICE commit!
LGTM

wpjunior (approved):

PR #2868, opened 2026-05-19
[chore] Always store unit ID in `app.run.shell` events

### LINKED ISSUE #2866: Record the effective pod/unit name in `tsuru app shell` events
**Is your feature request related to a problem? Please describe.**

`tsuru app shell` already creates an `app.run.shell` event and stores request metadata plus the shell command log. However, when the command runs without an explicit `unit` parameter, the event does not show which app unit actually received the shell session.

This makes debugging harder for apps with multiple units. In the Kubernetes provisioner, the tsuru unit name matches the pod name, so exposing the effective unit name in the event is enough to identify the pod that ran the shell.

**Describe the solution you'd like**

Update the shell flow so the event records the effective unit name used for the shell session.

For the initial scope, handle the regular unit-backed shell flow first:

- when `unit` is explicitly selected, persist that unit name in the event;
- when `unit` is omitted, persist the unit automatically selected for the session.

The result should be visible through `tsuru event info` for `app.run.shell` events.

**Describe alternatives you've considered**

- Relying only on the existing `unit` request parameter is not enough, because it is absent when the shell target is selected automatically.
- Including `--isolated` shell support in the same change is possible, but it likely needs extra plumbing from the Kubernetes provisioner because the isolated pod is created deeper in that path. That can be handled in a follow-up issue.

**Additional context**

Suggested implementation area:

- `api/shell.go`: resolve the effective unit before creating the event and include it in the event custom data;
- `api/shell_test.go`: add coverage for both default unit selection and explicit unit selection;
- event assertions may need to be extended wherever shell event info is validated.

**Acceptance criteria**

- `app.run.shell` events include the effective unit/pod name for regular shell sessions;
- the behavior works both with and without an explicit `unit` request parameter;
- the existing shell event log and metadata remain intact;
- automated tests cover the new behavior;
- `tsuru app shell --isolated` is explicitly documented as out of scope or follow-up for this issue.

marciosobel:
When running a shell with `tsuru shell`, the automatically chosen ID is also included in the event info. If a unit is provided, the provided ID will be prioritized.

closes #2866

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -89,6 +89,7 @@ Magno Torres <magno@corp.globo.com> <magnotorres@gmail.com>
 Manoel Domingues Junior <manoel.domingues.junior@gmail.com> <manoel.junior@corp.globo.com>
 Marc Abramowitz <msabramo@gmail.com> <marc@marc-abramowitz.com>
 Marcelo Jorge Vieira <metal@alucinados.com>
+Márcio Sobel <contact@marciosobel.dev>
 Marco Paulo Ollivier <marcopollivier@gmail.com>
 Marco Sinhoreli <msinhore@gmail.com>
 Marcus Cartágenes <mvcartagenes@gmail.com>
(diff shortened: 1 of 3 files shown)


ravilock:
@marciosobel can you add yourself to the CONTRIBUTORS file?

marciosobel:
@ravilock done!

ravilock (approved):

PR #2871, opened 2026-05-25
fix(net): remove inherited deadlines from WithoutCancel

immanuwell:
## Summary

`net.WithoutCancel` was still carrying the parent deadline.
So rollback and cleanup paths could still time out after switching to a non-cancelable ctx. Kinda defeats the point.

This swaps it to `context.WithoutCancel(...)` and keeps `CancelableParentContext(...)` working.

## Repro

1. `ctx, cancel := context.WithTimeout(context.Background(), 10*time.Millisecond)`
2. `ctx = tsuruNet.WithoutCancel(ctx)`
3. Before this patch, `ctx.Deadline()` still returned `ok=true`.
4. That lets rollback and deferred pod cleanup still hit the original timeout.

## Checks

- `go test ./net -count=1`
- `go test -c ./api`
- `go test -c ./action`
- `go test -c ./provision/servicecommon`
- `go test -c ./provision/kubernetes`

Mongo-backed package tests were not runnable in this local env

### DIFF
--- a/net/context.go
+++ b/net/context.go
@@ -20,18 +20,6 @@ func CancelableParentContext(ctx context.Context) context.Context {
 	return ctx
 }
 
-type withoutCancelContext struct {
-	context.Context
-}
-
-func (*withoutCancelContext) Err() error {
-	return nil
-}
-
-func (*withoutCancelContext) Done() <-chan struct{} {
-	return nil
-}
-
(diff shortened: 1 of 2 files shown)


wpjunior:
Thanks so very much @immanuwell

PR #2878, opened 2026-07-09
Feature/service fine grained permissions

ravilock:


api/permission.go
```diff
@@ -451,13 +451,11 @@ func listPermissions(w http.ResponseWriter, r *http.Request, t auth.Token) error
...
 
 type dynamicPermissionAction struct {
-	Name       string   `json:"name"`
-	Action     string   `json:"action"`
-	Method     string   `json:"method"`
-	Path       string   `json:"path"`
-	Scope      string   `json:"scope"`
-	EntityType string   `json:"entity_type"`
-	Contexts   []string `json:"contexts"`
+	Name     string   `json:"name"`
```
wpjunior: @ravilock Name field still here

ravilock: outdated hehe

VitinhoCodigoMarromeno:
Ficou Filé viu

permission/role.go
```diff
@@ -17,11 +18,12 @@ import (
 )
 
 type Role struct {
-	Name        string                `bson:"_id" json:"name"`
-	ContextType permTypes.ContextType `json:"context"`
-	Description string
-	SchemeNames []string `json:"scheme_names,omitempty"`
```
wpjunior, line 23: @ravilock we could simplify and also add dynamic permissions on SchemeNames ?

there is a reason to put in another slice?

esdrasbmXX:
Muito bom, parabéns!

ItaloJamison:
Ficou primeira de luxo

permission/role.go
ravilock: The main reason is trying to decrease blast radius on bugs.
There are two main things to note:
1. The static permissions depend on a global variable `PermissionRegistry`. This registry is only updated when tsuru api boots. It also has some generation logic and many authorization and role management paths depend on it
2. By separating static and dynamic permission into two separate fields, we do not need to change this PermissionRegistry code. I can also avoid introducing bugs and authorization issues on the main paths. A bug on dynamic permissions only affects service instance proxy routes. A bug on static permissoins might affect everything

wpjunior: Ok, understood =) great mitigation

api/permission.go
```diff
@@ -433,22 +434,167 @@ func listPermissions(w http.ResponseWriter, r *http.Request, t auth.Token) error
...
+// path: /roles/{name}/dynamic-permissions
+// method: POST
+// consume: application/x-www-form-urlencoded
+// responses:
+//
+//	200: Ok
+//	400: Invalid data
+//	401: Unauthorized
+//	409: Permission not allowed
+func addDynamicPermissions(w http.ResponseWriter, r *http.Request, t auth.Token) (err error) {
```
wpjunior: @ravilock can we reuse the actual add route, inferring whether is dynamic by prefix?

also can se put the prefix on a public constant ?

api/permission.go
```diff
@@ -433,22 +434,167 @@ func listPermissions(w http.ResponseWriter, r *http.Request, t auth.Token) error
...
+
+// title: remove dynamic permission
+// path: /roles/{name}/dynamic-permissions/{permission}
+// method: DELETE
+// responses:
+//
+//	200: Permission removed
+//	401: Unauthorized
+//	404: Not found
+func removeDynamicPermissions(w http.ResponseWriter, r *http.Request, t auth.Token) (err error) {
```
wpjunior: The same for here

api/permission.go
ravilock: Yeah, we are basically deciding here that all permissions with the `service-action` prefix are going to be considered public. Sounds like a reasonable decision.

wpjunior: yeap, also will reduce effort on tsuru-client side

permission/dynamic.go
```diff
@@ -5,67 +5,89 @@
...
-	return DynamicPermissionRegistry.list()
+// ExistsDynamic reports whether name is an ancestor-or-equal of an action
+// declared by an enabled service manifest.
+func ExistsDynamic(ctx context.Context, name string) (bool, error) {
+	serviceName, _, _ := strings.Cut(strings.TrimPrefix(name, DynamicPermissionPrefix), ".")
+	servicesCollection, err := storagev2.ServicesCollection()
+	if err != nil {
+		return false, err
+	}
+	var svc struct {
```
ravilock: I'd prefer to avoid declaring structs like these that are only scoped to a single function. The manifest is not that large in terms of memory to declare a different struct to reduce it.

wpjunior: me too, Mr Fable made this, lets move ServiceManifest to the types.

api/permission.go
```diff
@@ -433,18 +451,57 @@ func listPermissions(w http.ResponseWriter, r *http.Request, t auth.Token) error
...
 		for j, ctx := range contexts {
 			contextNames[j] = string(ctx)
 		}
-		permList[i] = permissionSchemeData{
+		permList = append(permList, permissionSchemeData{
 			Name:     perm.FullName(),
 			Contexts: contextNames,
+		})
+	}
+	services, err := service.GetServices(ctx)
```
wpjunior, line 479: @ravilock from line 466 to 503 we could extract to a method or a fuction

service/manifest.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+	removedActions := removedManifestActions(s.Manifest, manifest)
+	conflicts, err := manifestGrantConflicts(ctx, s.Name, removedActions, manifest)
+	if err != nil {
+		return err
+	}
+	if len(conflicts) > 0 && !force {
+		return &ManifestConflictError{Service: s.Name, Conflicts: conflicts}
+	}
+	if len(conflicts) > 0 {
```
crgisch, line 76: as we talked in the call, allowing the ingest with conflicts leaves orphaned grants behind
force=true logs and persists the manifest, it doesnt prune the flagged dynamic_scheme_names from the roles. They would still show in role info, and if the same action name is re-added later the stale grant silently reactivates

crgisch (approved):
LGTM 👍 
just the force ingest to take a look at

service/manifest.go
ravilock: We are going to just refuse to update the manifest if it will leave orphaned roles then

PR #2885, opened 2026-07-30
fix(log): don't log expected conditions as errors

### LINKED ISSUE #2884: Healthy happy paths logged at ERROR level: disabled job informer, platform "" image fallback, version 0 on first deploy
On a fresh install (observed on 1.30.2), three ERROR-level log lines fire on
completely healthy operations, which trains operators to ignore ERROR logs:

1. **Disabled job informer logged as ERROR + stack trace at startup.**
   When the cluster config deliberately sets `job-event-creation: false` (the
   default), `startJobInformer` returns an error and the caller logs it at
   ERROR with a full stack on every cluster controller startup:

   ```
   ERROR: error while starting job informer: job event creation is not enabled
   github.com/tsuru/tsuru/provision/kubernetes.(*clusterController).startJobInformer
       .../provision/kubernetes/monitor.go:141
   ...
   ```

   A feature that is configured off is not an error.

2. **Every container-image deploy logs `ERROR: Couldn't find images for
   platform ""`.** For platform-less deploys there is no platform image by
   definition, and the code continues with "fallback to default image name" by
   design (`app/image/platform.go`).

3. **Every first deploy of an app logs `ERROR: unable to find version 0 for
   app "x": Invalid version: 0`.** No versions exist before the first deploy —
   expected state, and the code already treats `InvalidVersionError` as
   non-fatal (`provision/servicecommon/actions.go`).

Operators triage ERROR-level logs; entries that fire on every healthy
deploy/startup bury real errors. Proposed: return `nil` for the
deliberately-disabled informer (a config choice is not an error) and downgrade
the other two to DEBUG. PR incoming.

guilhermebr:
Fixes #2884.

Three conditions that occur on healthy happy paths were logged at ERROR level
(one with a full stack trace). Operators triage ERROR logs; entries that fire
on every healthy deploy or startup train people to ignore them.

- **`provision/kubernetes/monitor.go`** — `startJobInformer` returned an error
  when `job-event-creation` is deliberately disabled in the cluster config,
  and the caller logs it at ERROR with a stack on every cluster controller
  startup. A configured-off feature is not an error: return `nil` (a DEBUG
  line notes the informer is skipped). Real informer startup failures are
  still returned and logged as before.
- **`app/image/platform.go`** — `CurrentImage` logged ERROR before falling
  back to the default image name, which is the designed behavior for every
  container-image (platform `""`) deploy: downgraded to DEBUG.
- **`provision/servicecommon/actions.go`** — `RunServicePipeline` logged ERROR
  for version 0 on every first deploy of an app, where no previous version can
  exist; the code already distinguishes real lookup failures
  (non-`InvalidVersionError` still returns an error): downgraded to DEBUG.

Tests (GoCheck, matching the existing suites): `startJobInformer` returns no
error when the feature is disabled; the other two assert via a captured logger
that the messages are emitted at DEBUG and nothing is logged at ERROR.

All three observed on every app deploy / startup of a healthy 1.30.2
installation.

### DIFF
--- a/provision/servicecommon/actions_test.go
+++ b/provision/servicecommon/actions_test.go
@@ -5,9 +5,11 @@
 package servicecommon
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"sort"
+	"strings"
 	"testing"
 
 	"github.com/pkg/errors"
@@ -16,6 +18,7 @@ import (
 	"github.com/tsuru/tsuru/app"
 	"github.com/tsuru/tsuru/app/version"
 	"github.com/tsuru/tsuru/db/storagev2"
+	"github.com/tsuru/tsuru/log"
 	"github.com/tsuru/tsuru/provision"
 	"github.com/tsuru/tsuru/provision/provisiontest"
 	"github.com/tsuru/tsuru/servicemanager"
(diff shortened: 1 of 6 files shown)


provision/servicecommon/actions_test.go
```diff
@@ -783,3 +786,23 @@ func (s *S) TestRunServicePipelineUpdateStates(c *check.C) {
...
+			"web": "python web1",
+		},
+	})
+	err := RunServicePipeline(context.TODO(), m, 0, provision.DeployArgs{
+		App:     fakeApp,
+		Version: newVersion,
+	}, nil)
+	c.Assert(err, check.IsNil)
+	c.Assert(strings.Contains(buf.String(), "unable to find version"), check.Equals, true)
+	c.Assert(strings.Contains(buf.String(), "ERROR"), check.Equals, false)
```
ravilock, line 807: shouldn't this value here be `DEBUG`?
If this test pass I guess that the `buf` is storing other logs and not only the one you are targeting to test on: https://github.com/tsuru/tsuru/pull/2885/changes#diff-a854476213bf7db7e9bf20d8109a11beea1b3dbe34e6c428fe81f17745b66b93R83

guilhermebr: It's asserting that it doesn't have the string "ERROR":
`"ERROR", check.Equals, false`

ravilock: Oh, now I see that..
Geez, this old assertion library we use is not that readable :/

ravilock (approved):

PR #2905, opened 2026-08-30
modernize: adopt slices, maps, min and range-over-int

guilhermebr:
Second slice of the Go 1.26 modernizer sweep, after #2904. This one covers the loop and collection fixers: `slicescontains`, `mapsloop`, `rangeint`, `minmax`, `forvar` and `stditerators`.

Generated with `go fix -<fixer> ./...`, one fixer at a time.

Unlike the `any` slice this is not a pure token swap, so it's worth an actual read — but it's small, and it removes 105 lines net.

### What changed

| Rewrite | Count | Example |
|---|---:|---|
| `slices.Contains` / `ContainsFunc` | 23 | hand-rolled `for … if v == x { return true }` |
| `maps.Copy` | 20 | `for k, v := range src { dst[k] = v }` |
| range-over-int | ~45 | `for i := 0; i < n; i++` → `for i := range n` |
| `min(…)` | 1 | an if-clamp on a slice bound |
| `stditerators` | 1 | `t.NumField()` index loop → `range t.Fields()` |

Only two imports are introduced: `slices` (19 files) and `maps` (10). No function signatures change.

### Review notes

Every `maps.Copy` replaced an unconditional copy loop — none had a filter or early exit that could have been silently folded away. The `slices.Contains` rewrites include two slightly awkward but faithful ones:

```go
hasName := slices.Contains(append(addrs, app.CName...), name)
if slices.ContainsFunc(teams, teamConstraint.check) { … }
```

All range-over-int rewrites have literal or loop-invariant bounds.

### One thing to know: the fixers are order-dependent

`stditerators` rewrites this reflect loop, and leaves a redundant self-shadow behind:

```go
 	fields := map[string]struct{}{}
-	for i := 0; i < t.NumField(); i++ {
-		field := t.Field(i)
+	for field := range t.Fields() {
+		field := field
```

`field := field` is dead under Go 1.22+ per-iteration loop variables — and it is precisely what `forvar` deletes, except `forvar` had already run by then. Re-running `forvar` after `stditerators` removes it, which is what this branch does. Worth knowing if you ever run the fixers yourself: a single pass can leave residue that a second pass cleans up.

### Verified

- `go build ./...`, `go vet ./...`, `gofmt -l .` — clean
- `golangci-lint run ./...` (v2.11.4) — 0 issues
- full unit suite, 64 packages — pass
- `tsurud api --dry` — pass

No generated code is touched.


wpjunior:
Hi @guilhermebr there is a way to avoid regression about these modernizations? To ensure the new code always use these new golang standards

guilhermebr:
@wpjunior Sure, I just committed this change to CI for new code only (if enabled for the entire codebase, it will fail until all PRs that modernize the code are merged). After merging all PRs (there are more coming), we can switch to checking the entire codebase.

wpjunior (approved):

ravilock (approved):

PR #2911, opened 2026-09-03
Fix/cluster storage breaking kubeconfig

ravilock:

PR #2913, opened 2026-09-03
Fix/cluster storage breaking kubeconfig (#2911)

ravilock:
This PR merges the commit that fixes the cluster config persistence and integration testing

* chore: check which docker compose tool is available on user setup

* test: cluster config write operations should preserve valid kube-config

* fix: provisioner cluster collection should not convert nil slices/maps to empty on storage writes

* test: extend retry time for job completion check on integration

* test: describe broken pods after failed integration testing

* fix: nil map should be read as empty for provisioner cluster

* chore: setup minikube and debug registry-insecure reads

* chore: pin minikube version

* test: use testify require on regression test

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -46,6 +46,9 @@ jobs:
       uses: medyagh/setup-minikube@master
       with:
         kubernetes-version: "v${{ matrix.k8s_version }}"
+        container-runtime: docker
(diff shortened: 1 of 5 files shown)


devxcelio:
Interesting

wpjunior:
amazing @ravilock

wpjunior (approved):

PR #2917, opened 2026-09-15
expect tsuru team info command to work with different role contexts

ravilock:
…d perm team read with contexts


wpjunior (approved):
nice catch @ravilock

crgisch (approved):
LGTM 🚀

